PR #16, opened 2024-05-30
feat(build): Add Dockerfile

evanpurkhiser:


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+WORKDIR /app
+
+# Just copy the Cargo.toml files and trigger a build so that we compile our
+# dependencies only. This way we avoid layer cache invalidation if our
+# dependencies haven't changed, resulting in faster builds.
+
+COPY Cargo.toml .
+COPY Cargo.lock .
+RUN cargo build --release && rm -rf src/
+RUN strip target/release/uptime-checker
```
joshuarli, line 19: https://doc.rust-lang.org/cargo/reference/profiles.html#strip

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM alpine:3.18
+
+COPY --from=builder /app/target/release/uptime-checker /usr/local/bin/uptime-checker
+
+RUN apk add --no-cache tini
+
+RUN addgroup -S app && adduser -S app -G app
+USER app
+
+ENTRYPOINT ["/sbin/tini", "--", "uptime-checker"]
```
joshuarli, line 35:
```suggestion
ENTRYPOINT ["/sbin/tini", "--", "/usr/local/bin/uptime-checker"]
```

may as well save a path lookup and be absolutely sure you're calling the right one

Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM rust:1.74-alpine3.18 as builder
+
+RUN mkdir -p ~/.cargo && \
+    echo '[registries.crates-io]' > ~/.cargo/config && \
+    echo 'protocol = "sparse"' >> ~/.cargo/config
+
+RUN apk add --no-cache libc-dev
+
+RUN USER=root cargo new --bin /app
```
IanWoodard, line 9: Could we create and use a non-root user? That way we follow the principle of least privilege.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY Cargo.lock .
+RUN cargo build --release && rm -rf src/
+RUN strip target/release/uptime-checker
+
+# Copy the source code and run the build again. This should only compile the
+# app itself as the dependencies were already built above.
+COPY . ./
+RUN rm target/release/deps/uptime_checker* && cargo build --release
+
+FROM alpine:3.18
```
joshuarli, line 26: 3.20 is available btw

Dockerfile
IanWoodard: Fair point

evanpurkhiser:
Good for a second look

joshuarli (approved):

PR #17, opened 2024-05-30
feat(build): Add cloudbuild.yaml

evanpurkhiser:


cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+        '.',
+      ]
+    env: [DOCKER_BUILDKIT=1]
+
+  - name: 'gcr.io/cloud-builders/docker'
+    entrypoint: 'bash'
+    args:
+      - '-c'
+      - |
+        # Only push "latest" tag when building on "master"
```
joshuarli, line 23: main, not master in this case

evanpurkhiser: good catch

PR #20, opened 2024-05-31
feat(deploy): Add gocd pipedream pipelines

evanpurkhiser:
Deploys the uptime-checker via gocd

### DIFF
--- a/.github/workflows/validate-pipelines.yml
+++ b/.github/workflows/validate-pipelines.yml
@@ -1,0 +1,60 @@
+name: Validate Deployment Pipelines
+
+on:
+  pull_request:
+  push:
+      branches: [main, test-me-*]
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+    cancel-in-progress: true
+
+jobs:
+    files-changed:
+        name: files-changed
+        runs-on: ubuntu-latest
+        # Map a step output to a job output
+        outputs:
+            gocd: ${{ steps.changes.outputs.gocd }}
+        steps:
+          - uses: actions/checkout@v3
+          - name: Check for relevant file changes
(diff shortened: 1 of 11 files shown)


.github/workflows/validate-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+
+jobs:
+    files-changed:
+        name: files-changed
+        runs-on: ubuntu-latest
+        # Map a step output to a job output
+        outputs:
+            gocd: ${{ steps.changes.outputs.gocd }}
+        steps:
+          - uses: actions/checkout@v3
```
IanWoodard, line 20: Ideally we should use shas rather than versions. Also, in this case v4 is newer and should be used instead of v3.

.github/workflows/validate-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+        name: Validate GoCD Pipelines
+        runs-on: ubuntu-latest
+
+        # required for google auth
+        permissions:
+            contents: "read"
+            id-token: "write"
+
+        steps:
+            - uses: actions/checkout@v3
```
IanWoodard, line 42: Same as above

.github/workflows/validate-pipelines.yml
```diff
@@ -0,0 +1,60 @@
...
+            - id: 'auth'
+              uses: google-github-actions/auth@v1
```
IanWoodard, line 44: Same as above, there is a newer version available as well.

.github/workflows/validate-pipelines.yml
evanpurkhiser: Just copied from your PR 😏

IanWoodard (approved):

PR #21, opened 2024-06-03
feat(scheduler): Add simple tokio-cron-scheduler job

evanpurkhiser:
Just an example of what our celery-beat style task might look like

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1,0 +1,14 @@
+use tokio_cron_scheduler::{Job, JobScheduler, JobSchedulerError};
+
+pub async fn run_scheduler() -> Result<(), JobSchedulerError> {
+    let scheduler = JobScheduler::new().await?;
+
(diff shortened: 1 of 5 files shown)


src/scheduler.rs
```diff
@@ -0,0 +1,14 @@
+use tokio_cron_scheduler::{Job, JobScheduler, JobSchedulerError};
+
+pub async fn run_scheduler() -> Result<(), JobSchedulerError> {
+    let scheduler = JobScheduler::new().await?;
+
+    let checker_job = Job::new("1/10 * * * * *", |_uuid, _l| {
```
lynnagara, line 6: what does `1/10 * * * * *` mean?

evanpurkhiser: good question, just copied this from their docs. It's once every 10 seconds

lynnagara: ah thanks, sorry wasn't familiar with that format

wedamija (approved):

PR #22, opened 2024-06-04
feat: Add function to perform domain checks

wedamija:
Add initial function to perform domain checks. We need better testing around this and probably also need return more useful failure reasons.

I'll follow up later with a pr that adds mocking for the domains so we can test all the cases.

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -33,8 +33,6 @@ pub async fn do_request(
 /// Makes a request to a url to determine whether it is up.
 /// Up is defined as returning a 2xx within a specific timeframe.
 pub async fn check_url(client: &reqwest::Client, url: String) -> CheckResult {
-
-
     match do_request(client, url).await {
         Ok(_) => CheckResult::SUCCESS,
         Err(e) => {
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -15,29 +15,27 @@ pub enum FailureReason {
     DnsError(String),
     Error(String),
 }
-
+/// Fetches the response from a URL.
+///
+/// First attempts to fetch just the head, and if not supported falls back to fetching the entire body.
 pub async fn do_request(
     client: &reqwest::Client,
-    domain: String,
+    url: String,
 ) -> Result<Response, reqwest::Error> {
-    /*
-    Fetches the response from a URL. First attempts to fetch just the head, and if not supported falls
-    back to fetching the entire body.
-     */
-    let response = client.head(&domain).send().await?;
+    let response = client.head(&url).send().await?;
     match response.status() {
-        StatusCode::METHOD_NOT_ALLOWED => client.get(&domain).send().await,
+        StatusCode::METHOD_NOT_ALLOWED => client.get(&url).send().await,
         _ => Ok(response),
     }
 }
 
-pub async fn check_domain(client: &reqwest::Client, domain: String) -> CheckResult {
-    /*
-    Checks whether a domain is up and working. Makes a request to the domain and if a 2xx is
-    returned then the domain is considered working.
-     */
 
-    match do_request(client, domain).await {
+/// Makes a request to a url to determine whether it is up.
+/// Up is defined as returning a 2xx within a specific timeframe.
+pub async fn check_url(client: &reqwest::Client, url: String) -> CheckResult {
+
+
+    match do_request(client, url).await {
         Ok(_) => CheckResult::SUCCESS,
         Err(e) => {
             if e.is_timeout() {
@@ -65,7 +63,7 @@ pub async fn check_domain(client: &reqwest::Client, domain: String) -> CheckResu
 
 #[cfg(test)]
 mod tests {
-    use super::{check_domain, CheckResult, FailureReason};
+    use super::{check_url, CheckResult, FailureReason};
     use httpmock::prelude::*;
     use httpmock::Method;
     use reqwest::{ClientBuilder, StatusCode};
@@ -96,15 +94,15 @@ mod tests {
         let client = ClientBuilder::new().timeout(timeout).build().unwrap();
 
         assert_eq!(
-            check_domain(&client, server.url("/head")).await,
+            check_url(&client, server.url("/head")).await,
             CheckResult::SUCCESS
         );
         assert_eq!(
-            check_domain(&client, server.url("/no-head")).await,
+            check_url(&client, server.url("/no-head")).await,
             CheckResult::SUCCESS
         );
         assert_eq!(
-            check_domain(&client, server.url("/timeout")).await,
+            check_url(&client, server.url("/timeout")).await,
             CheckResult::FAILURE(FailureReason::Timeout)
         );
         head_mock.assert();
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -1,7 +1,6 @@
 use std::error::Error;
-use std::time::Duration;
 
-use reqwest::{ClientBuilder, StatusCode};
+use reqwest::StatusCode;
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub enum CheckResult {
@@ -27,33 +26,28 @@ pub async fn check_domain(client: &reqwest::Client, domain: String) -> CheckResu
     }
 
     match response_result {
-        Ok(_) => {
-            return CheckResult::SUCCESS;
-        }
+        Ok(_) => CheckResult::SUCCESS,
         Err(e) => {
             if e.is_timeout() {
                 return CheckResult::FAILURE(FailureReason::Timeout);
-            } else {
-                // TODO: More reasons
-
-                let mut source = e.source();
-                while let Some(inner) = source {
-                    if inner.source().is_none() {
-                        break;
-                    }
-                    source = inner.source();
-
-                    // TODO: Would be better to get specific errors without string matching like this
-                    // Not sure if there's a better way
-                    if source.unwrap().to_string().contains("dns error") {
-                        return CheckResult::FAILURE(FailureReason::DnsError(
-                            source.unwrap().source().unwrap().to_string(),
-                        ));
-                    }
-                }
-
-                return CheckResult::FAILURE(FailureReason::Error(format!("{:?}", e)));
             }
+            // TODO: More reasons
+            let mut source = e.source();
+            while let Some(inner) = source {
+                if inner.source().is_none() {
+                    break;
+                }
+                source = inner.source();
+
+                // TODO: Would be better to get specific errors without string matching like this
+                // Not sure if there's a better way
+                if source.unwrap().to_string().contains("dns error") {
+                    return CheckResult::FAILURE(FailureReason::DnsError(
+                        source.unwrap().source().unwrap().to_string(),
+                    ));
+                }
+            }
+            CheckResult::FAILURE(FailureReason::Error(format!("{:?}", e)))
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    match response_result {
+        Ok(_) => {
+            return CheckResult::SUCCESS;
+        }
+        Err(e) => {
+            if e.is_timeout() {
+                return CheckResult::FAILURE(FailureReason::Timeout);
+            } else {
```
evanpurkhiser, line 46: no need for the else if you're returning

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+    Error(String),
+}
+
+pub async fn check_domain(domain: String) -> CheckResult {
+    // TODO: Timeout should come from config
+    let timeout = Duration::new(5, 0);
+    let client_result = ClientBuilder::new().timeout(timeout).build();
+    let client = match client_result {
+        Err(e) => {
+            return CheckResult::FAILURE(FailureReason::ConfigError(e.to_string()));
```
wedamija, line 27: Maybe this should be missed since it's a failure on our end?

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        Ok(client) => client,
+    };
```
evanpurkhiser, line 30: I think there's a more idiomatic way to do this

```suggestion
    let client = client_result.map_err(|e| CheckResult::FAILURE(FailureReason::ConfigError(e.to_string()))?
```

(you'll have to check me on this)

read into the the `?` operator, you'll see it a lot. Basically it unwraps `Result<>` types and early returns if an error comes back.

evanpurkhiser: Actually you can probably do this on the client itself

```rs
let client = ClientBuilder::new()
  .timeout(timeout)
  .build();
  .map_err(|e| CheckResult::FAILURE(FailureReason::ConfigError(e.to_string()))?
```

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+        if response.status() == StatusCode::METHOD_NOT_ALLOWED {
+            // If this domain doesn't accept HEAD, attempt GET request
+            response_result = client.get(&domain).send().await;
+        }
+    }
+
+    match response_result {
+        Ok(_) => {
+            return CheckResult::SUCCESS;
+        }
```
evanpurkhiser, line 42: Since this is the last expression you can just do this, it will be passed up
```suggestion
        Ok(_) => CheckResult::SUCCESS;
```

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+            CheckResult::SUCCESS
+        );
+        assert_eq!(
+            check_domain("https://sentry.io".to_string()).await,
+            CheckResult::SUCCESS
+        );
+        assert_eq!(check_domain("https://hjkhjkljkh.io".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
+        assert_eq!(check_domain("https://santry.io".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
+    }
+}
```
evanpurkhiser, line 88: This requires internet to run right? ☠️

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+    Timeout,
+    DnsError(String),
+    ConfigError(String),
+    Error(String),
+}
+
+pub async fn check_domain(domain: String) -> CheckResult {
+    // TODO: Timeout should come from config
+    let timeout = Duration::new(5, 0);
+    let client_result = ClientBuilder::new().timeout(timeout).build();
```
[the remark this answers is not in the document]
wedamija: Not sure if there are any disadvantages to keeping it around, I'll take a look

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+pub enum FailureReason {
+    Timeout,
+    DnsError(String),
+    ConfigError(String),
+    Error(String),
+}
+
+pub async fn check_domain(domain: String) -> CheckResult {
```
vartec, line 21: You're using URLs without trailing `/`, so you get 301 redirects, which Reqwest handles automatically, but it's kinda unnecessary.

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+                return CheckResult::FAILURE(FailureReason::Timeout);
+            } else {
+                // TODO: More reasons
+
+                let mut source = e.source();
+                while let Some(inner) = source {
+                    if inner.source().is_none() {
+                        break;
+                    }
+                    source = inner.source();
```
vartec, line 54: Rust noob question: wouldn't more idiomatic way be to not use `mut`, but instead shadow it with `let source = inner.source()` in line 54?

src/checker.rs
vartec: https://github.com/lipanski/mockito ?

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+pub enum FailureReason {
+    Timeout,
+    DnsError(String),
+    ConfigError(String),
+    Error(String),
+}
+
+pub async fn check_domain(domain: String) -> CheckResult {
+    // TODO: Timeout should come from config
+    let timeout = Duration::new(5, 0);
```
loewenheim: I think `from_secs` is clearer.

```suggestion
    let timeout = Duration::from_secs(5);
```

src/checker.rs
loewenheim: The problem with this is that `check_domain` returns a `CheckResult`, which is not the same thing as a `Result` (it has 3 variants to `Result'`'s 2). In general I would strongly prefer using `Result` if at all possible. In this case that could look like
```
enum CheckError {
    Missed,
    Failure(FailureReason),
}
```
`FailureReason` could be exactly as it is now and `CheckResult` could be an alias for `Result<(), CheckError>` (since the `Ok` case contains no information). Alternatively, one could also do
```
enum CheckError {
    Missed,
    Timeout,
    DnsError(String),
    ConfigError(String),
    Error(String),
}
```
and ditch `FailureReason` to remove a level of indirection.

src/checker.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        Ok(client) => client,
+    };
+    let mut response_result = client.head(&domain).send().await;
+    if let Ok(response) = &response_result {
+        if response.status() == StatusCode::METHOD_NOT_ALLOWED {
+            // If this domain doesn't accept HEAD, attempt GET request
+            response_result = client.get(&domain).send().await;
+        }
+    }
```
loewenheim, line 26: Immediately after this you're matching on `response_result` anyway. You can move this into the `Ok` branch.

src/checker.rs
loewenheim: If you use shadowing the loop doesn't work. You need to mutate the outer `source` variable that loop condition can see.

src/checker.rs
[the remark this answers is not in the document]
wedamija: Passing it through for now, although we should just instantiate this once when we call this function

src/checker.rs
wedamija: Yeah, I'm going to put mocks here later

src/checker.rs
wedamija: Yeah, this is nicer, and we don't need `source` outside the loop anyway. I think I had that in place from some earlier experiments

loewenheim: The entire loop can be improved like so:
```rust
                let mut inner = &e as &dyn Error;
                while let Some(source) = inner.source() {
                    inner = source;

                    // TODO: Would be better to get specific errors without string matching like this
                    // Not sure if there's a better way
                    let inner_message = inner.to_string();
                    if inner_message.contains("dns error") {
                        return CheckResult::FAILURE(FailureReason::DnsError(inner_message));
                    }
                }
```
I'll see if I can find anything about the string matching, that is indeed very unappealing.

loewenheim (changes requested):
In general I would recommend running `clippy` and setting it up as part of your CI. It's really helpful. Locally you can run it with `cargo clippy`.

src/checker.rs
loewenheim: Oh, sorry, I didn't catch that `response_result` was mutable. That works, but in my opinion it's a bit surprising. If you had
1. `CheckResult` as a conventional `Result` type (see above),
2. An impl of `From<reqwest::Error> for `CheckError`
this could look like
```rust
    let response = client.head(&domain).send().await?;
    if response.status() == StatusCode::METHOD_NOT_ALLOWED {
        client.get(&domain).send().await?;
    }
    Ok(())

File: src/checker.rs (hunk 3 above)
Comment (loewenheim) [reply src/checker.rs]: I see what you mean. I think you're correct that `CheckResult` should not be a `Result` after all, and that `Missed` should not be a variant of this enum. One simple way to accomplish what you wrote in pseudocode could be to use `Option<CheckResult>` as the type of `result`, with `None` meaning `Missed`.

File: src/checker.rs (hunk 8 above)
Comment (wedamija) [reply src/checker.rs]: Yeah, I'm planning on attempting to improve this as well, but so far it looks like reqwest doesn't provide a lot of info here. We can do a lot better with status codes at least, but for other network related issues it might be tricky

File: src/checker.rs [hunk 11]
```diff
@@ -0,0 +1,77 @@
...
+        assert_eq!(
+            check_domain(&client, "https://google.com".to_string()).await,
+            CheckResult::SUCCESS
+        );
+        assert_eq!(
+            check_domain(&client, "https://sentry.io".to_string()).await,
+            CheckResult::SUCCESS
+        );
+        assert_eq!(check_domain(&client, "https://hjkhjkljkh.io".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
+        assert_eq!(check_domain(&client, "https://santry.io".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
```
wedamija, line 75: These messages are different in ci, so expecting specific strings isn't a great idea. I'll look into other options

src/checker.rs
evanpurkhiser: This is discarding too much right. I think we would want

```rs
let response = client.head(&domain).send().await?;
match response.status() {
  StatusCode::METHOD_NOT_ALLOWED => client.get(&domain).send().await?
  _ => response
}
```

src/checker.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+pub async fn do_request(
+    client: &reqwest::Client,
+    domain: String,
+) -> Result<Response, reqwest::Error> {
+    /*
+    Fetches the response from a URL. First attempts to fetch just the head, and if not supported falls
+    back to fetching the entire body.
+     */
```
loewenheim, line 26: If these are intended as doc comments, they should go above the function with triple slashes, like so:
```suggestion
/// Fetches the response from a URL.
///
/// First attempts to fetch just the head, and if not supported falls
/// back to fetching the entire body.
pub async fn do_request(
    client: &reqwest::Client,
    domain: String,
) -> Result<Response, reqwest::Error> {
```

evanpurkhiser (approved):

PR #23, opened 2024-06-05
feat: Add result types

evanpurkhiser:
This matches what we've defined in the kafka schemas


wedamija (approved):
lgtm, although were we going to use msgpack instead?

PR #28, opened 2024-06-06
feat: Execute checker every 5 minutes on sentry.io

evanpurkhiser:
Actually does something now

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1,10 +1,29 @@
+use std::time::Duration;
+
+use chrono::Utc;
+use reqwest::ClientBuilder;
 use tokio_cron_scheduler::{Job, JobScheduler, JobSchedulerError};
+
+use crate::checker::check_url;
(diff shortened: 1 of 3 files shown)


src/scheduler.rs
```diff
@@ -1,10 +1,29 @@
...
+
 pub async fn run_scheduler() -> Result<(), JobSchedulerError> {
     let scheduler = JobScheduler::new().await?;
 
-    let checker_job = Job::new("1/10 * * * * *", |_uuid, _l| {
-        println!("I run every 10 seconds");
+    let client = ClientBuilder::new()
+        .timeout(Duration::from_secs(5))
+        .build()
+        .expect("Failed to build checker client");
```
evanpurkhiser, line 15: We'll probably want a `Checker::new()` where it does this client building thing, but this is fine for now

wedamija (approved):

PR #32, opened 2024-06-07
feat: Return types::CheckResult from check_url

evanpurkhiser:


src/checker.rs
```diff
@@ -15,62 +21,162 @@ pub enum FailureReason {
...
 /// First attempts to fetch just the head, and if not supported falls back to fetching the entire body.
-async fn do_request(client: &reqwest::Client, url: &str) -> Result<Response, reqwest::Error> {
-    let response = client.head(url).send().await?;
-    match response.status() {
-        StatusCode::METHOD_NOT_ALLOWED => client.get(url).send().await,
-        _ => Ok(response),
+async fn do_request(
+    client: &reqwest::Client,
+    url: &str,
+) -> Result<(RequestType, Response), (RequestType, reqwest::Error)> {
```
evanpurkhiser, line 31: I don't totally love this

loewenheim: I would change this to
```suggestion
) -> (RequestType, Result<Response, reqwest::Error>) {
```
You can't use `?` in the function anymore, but it's not too bad:
```rust
    let response = match client.head(url).send().await {
        Ok(response) => response,
        Err(e) => return (RequestType::Head, Err(e)),
    };

    // Successful head request
    if response.status() != StatusCode::METHOD_NOT_ALLOWED {
        return (RequestType::Head, Ok(response));
    }

    let result = client.get(url).send().await;

    (RequestType::Get, result)
```

src/checker.rs
```diff
@@ -15,62 +21,162 @@ pub enum FailureReason {
...
+            // TODO: Would be better to get specific errors without string matching like this
+            // Not sure if there's a better way
+            let inner_message = inner.to_string();
+            if inner_message.contains("dns error") {
+                return Some(inner.source().unwrap().to_string());
+            }
+        }
+        None
     }
 }
```
loewenheim, line 71: No reason to introduce a trait for this, this can just be a function `is_dns_error(&reqwest::Error) -> Option<String>.`

src/checker.rs
evanpurkhiser: Yeah agree that does feel better

src/checker.rs
```diff
@@ -15,62 +21,153 @@ pub enum FailureReason {
...
+    let trace_id = Uuid::new_v4();
+
+    let start = Instant::now();
+    let (request_type, response) = do_request(client, url).await;
+    let duration_ms = Some(start.elapsed().as_millis());
+
+    let status = match response.as_ref().map_or(false, |r| r.status().is_success()) {
+        true => CheckStatus::Success,
+        false => CheckStatus::Failure,
+    };
```
loewenheim, line 73: `match` with `true` and `false` is kind of unusual; you'd more commonly use `if-else`. It's fine as it is though.

src/checker.rs
```diff
@@ -15,62 +21,153 @@ pub enum FailureReason {
...
+
+    let http_status_code = response
+        .as_ref()
+        .map(|r| Some(r.status()))
+        .unwrap_or_else(|e| e.status())
+        .map(|s| s.as_u16());
```
loewenheim, line 79: I think this is clearer as a `match`.
```suggestion
    let http_status_code = match &response {
        Ok(r) => Some(r.status().as_u16()),
        Err(e) => e.status().map(|s| s.as_u16()),
    };
```

src/checker.rs
```diff
@@ -1,9 +1,15 @@
...
+use tokio::time::Instant;
+use uuid::Uuid;
+
+use crate::types::{
+    CheckResult, CheckStatus, CheckStatusReason, CheckStatusReasonType, RequestInfo, RequestType,
+};
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
-pub enum CheckResult {
+pub enum OldResult {
```
loewenheim, line 12: Is this type even still used anywhere?

src/checker.rs
```diff
@@ -15,62 +21,153 @@ pub enum FailureReason {
...
+    });
+
+    let status_reason = match response {
+        Ok(r) => match r.status().is_success() {
+            true => None,
+            false => Some(CheckStatusReason {
+                status_type: CheckStatusReasonType::Failure,
+                description: format!("Got non 2xx status: {}", r.status()),
+            }),
+        },
```
loewenheim, line 93: Here I would suggest a match guard:
```suggestion
        Ok(r) if !r.status().is_success() => Some(CheckStatusReason {
            status_type: CheckStatusReasonType::Failure,
            description: format!("Got non 2xx status: {}", r.status()),
        }),
        Ok(_) => None,
```
Or alternatively
```suggestion
        Ok(r) if r.status().is_success() => None,
        Ok(r) => Some(CheckStatusReason {
            status_type: CheckStatusReasonType::Failure,
            description: format!("Got non 2xx status: {}", r.status()),
        }),
```
if you don't want to have the negated conditional.

loewenheim (commented):
Got some more stylistic remarks/nitpicks.

src/checker.rs
evanpurkhiser: oh nope, needs removed 👍

src/checker.rs
evanpurkhiser: Ohhh that's clean

src/checker.rs
evanpurkhiser: ah yeah that makes sense

src/checker.rs
evanpurkhiser: I didn't know about match guards thank you

src/checker.rs
```diff
@@ -1,76 +1,157 @@
...
+
+    let start = Instant::now();
+    let (request_type, response) = do_request(client, url).await;
+    let duration_ms = Some(start.elapsed().as_millis());
+
+    let status = if response.as_ref().map_or(false, |r| r.status().is_success()) {
+        CheckStatus::Success
+    } else {
+        CheckStatus::Failure
+    };
```
loewenheim, line 60: I just remembered there's a nice new API for this that I've never used myself:
```suggestion
    let status = response.is_ok_and(|r| r.status().is_success());
```

src/checker.rs
```diff
@@ -1,76 +1,157 @@
...
-    match do_request(client, url).await {
-        Ok(_) => CheckResult::Success,
-        Err(e) => {
+    let trace_id = Uuid::new_v4();
+
+    let start = Instant::now();
+    let (request_type, response) = do_request(client, url).await;
+    let duration_ms = Some(start.elapsed().as_millis());
+
+    let status = if response.as_ref().map_or(false, |r| r.status().is_success()) {
```
loewenheim, line 56: I just remembered that there is a nice new API for this that I've never used myself:
```suggestion
    let status = if response.is_ok_and(|r| r.status().is_success()) {
```

evanpurkhiser: ah perfect!

evanpurkhiser: Still wants the `as_ref()`

loewenheim: Yeah that was my bad.

wedamija (approved):

PR #33, opened 2024-06-07
feat: Implement basic producer

wedamija:
This implements the basics of a producer for sending results back to sentry.

### DIFF
--- a/src/producer.rs
+++ b/src/producer.rs
@@ -17,29 +17,38 @@ pub enum ExtractCodeError {
     ProducerError(#[from] ProducerError),
 }
 
-pub async fn produce_checker_result(
-    result: &CheckResult,
-    topic: Topic,
-    config: KafkaConfig,
-) -> Result<(), ExtractCodeError> {
-    let producer = KafkaProducer::new(config.clone());
-    let topic = TopicOrPartition::Topic(topic);
-    // TODO: Note that this will rarely fail directly here - mostly will only happen if the queue
-    // is full. We need to have a callback that tells us whether the other produces succeed so we
-    // know whether to mark the check as successful. We never want to just silently drop a result -
-    // we'd much prefer to retry it later and produce a miss, so that we at least know that we missed
-    // and don't have holes.
-    Ok(producer.produce(
-        &topic,
-        KafkaPayload::new(None, None, Some(serde_json::to_vec(result)?)),
-    )?)
+pub struct ResultProducer {
+    producer: KafkaProducer,
+    topic: TopicOrPartition,
+}
+
+impl ResultProducer {
+    pub fn new(topic_name: &str, config: KafkaConfig) -> Self {
+        let producer = KafkaProducer::new(config);
+        let topic = TopicOrPartition::Topic(Topic::new(topic_name));
+        Self { producer, topic }
+    }
+
+    pub async fn produce_checker_result(
+        &self,
+        result: &CheckResult,
+    ) -> Result<(), ExtractCodeError> {
+        // TODO: Note that this will rarely fail directly here - mostly will only happen if the queue
+        // is full. We need to have a callback that tells us whether the other produces succeed so we
+        // know whether to mark the check as successful. We never want to just silently drop a result -
+        // we'd much prefer to retry it later and produce a miss, so that we at least know that we missed
+        // and don't have holes.
+        Ok(self.producer.produce(
+            &self.topic,
+            KafkaPayload::new(None, None, Some(serde_json::to_vec(result)?)),
+        )?)
+    }
 }
 
 #[cfg(test)]
 mod tests {
-    use super::{ExtractCodeError, produce_checker_result};
+    use super::{ExtractCodeError, ResultProducer};
     use rust_arroyo::backends::kafka::config::KafkaConfig;
-    use rust_arroyo::types::Topic;
     use uuid::Uuid;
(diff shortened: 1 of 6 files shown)


src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[error(transparent)]
+    SerializationError(#[from] serde_json::Error),
+    #[error(transparent)]
+    ProducerError(#[from] ProducerError),
+}
+
+pub async fn produce_checker_result(
+    result: CheckResult,
+    topic: Topic,
+    config: KafkaConfig,
```
wedamija, line 23: It might make sense for this to not have to accept the topic/config, and for us to have it set up somewhere

evanpurkhiser: Yeah I agree, we can just put this method on a struct and load it in there

src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+    // and don't have holes.
+    Ok(producer.produce(
+        &topic,
+        KafkaPayload::new(None, None, Some(serde_json::to_vec(&result)?)),
+    )?)
+}
+
+#[cfg(test)]
+mod tests {
+    use chrono::NaiveDateTime;
```
evanpurkhiser, line 40: Using this?

src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+    use chrono::offset::Local;
```
evanpurkhiser, line 41: Unused I think

src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+};
+use rust_arroyo::backends::{Producer, ProducerError};
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExtractCodeError {
+    #[error(transparent)]
+    SerializationError(#[from] serde_json::Error),
+    #[error(transparent)]
+    ProducerError(#[from] ProducerError),
+}
```
evanpurkhiser, line 18: Damn I like that

evanpurkhiser (approved):

src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+    use chrono::offset::Local;
+    use super::{ExtractCodeError, produce_checker_result};
+    use rust_arroyo::backends::kafka::config::KafkaConfig;
+    use rust_arroyo::types::Topic;
+    use uuid::Uuid;
+    use crate::checker::{check_url, FailureReason, UrlCheckResult};
+    use crate::checker::UrlCheckResult::Success;
+    use crate::types::{CheckResult, CheckStatus, CheckStatusReason, CheckStatusReasonType, RequestInfo, RequestType};
+
+    pub async fn send_result(result: UrlCheckResult) -> Result<(), ExtractCodeError>{
```
evanpurkhiser, line 50: `result` seems unused here

src/checker.rs
```diff
@@ -3,7 +3,7 @@ use std::error::Error;
 use reqwest::{Response, StatusCode};
 
 #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
-pub enum CheckResult {
+pub enum UrlCheckResult {
```
evanpurkhiser, line 6: Want to actually drop these changes from your patch? It might be easier if I merge mine first actually

src/producer.rs
wedamija: Oh yeah, that's some test junk I should remove

src/producer.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[error(transparent)]
+    ProducerError(#[from] ProducerError),
+}
+
+pub async fn produce_checker_result(
+    result: CheckResult,
+    topic: Topic,
+    config: KafkaConfig,
+) -> Result<(), ExtractCodeError> {
+    let producer = KafkaProducer::new(config.clone());
```
evanpurkhiser, line 25: Why is the clone needed here? doesn't look like we're using the config anywhere else?

src/producer.rs
```diff
@@ -0,0 +1,78 @@
...
+        produce_checker_result(&result, Topic::new("test-topic"), config).await
+    }
+
+    #[tokio::test]
+    async fn test() {
+        let result = send_result(CheckStatus::Success).await;
+        assert_eq!(result.unwrap(), ());
+    }
+
+}
```
evanpurkhiser, line 78: nit of all nits: trailing newline, I think rustfmt should catch no?

src/scheduler.rs
```diff
@@ -17,6 +20,11 @@ pub async fn run_scheduler() -> Result<(), JobSchedulerError> {
             println!("Executing job at {:?}", Utc::now());
 
             let check_result = job_checker.check_url("https://sentry.io").await;
+            // TODO: Get this from configuration.
+            // TODO: Producer should be instantiated with these values and shared
+            let config = KafkaConfig::new_config(["0.0.0.0".to_string()].to_vec(), None);
+            let topic = Topic::new("uptime-checker-results");
+            let _ = produce_checker_result(&check_result, topic, config).await;
```
evanpurkhiser, line 27: nice, let's move the config outside of the function though I think.

If you address the earlier comment of making a ResultsProducer struct we can pull that out also (will need to wrap in Arc)


Once #43 is merged we can pass the config into `run_scheduler` and it can construct the KafkaConfig from that

evanpurkhiser (approved):

PR #36, opened 2024-06-10
feat: Add sentry

evanpurkhiser:


src/cli.rs
```diff
@@ -2,15 +2,23 @@ use std::io;
 
 use tokio::signal::ctrl_c;
 
-use crate::{cliapp::make_app, scheduler::run_scheduler};
+use crate::{cliapp::make_app, logging, scheduler::run_scheduler};
 
-pub async fn execute() -> io::Result<()> {
```
evanpurkhiser, line 7: Per our documentation

> In a multithreaded application, spawned threads should inherit the Hub from the main thread. In order for that to happen, the Sentry client must be initialized before starting an async runtime or spawning threads. This means you'll have to avoid using macros such as #[tokio::main] or #[actix_web::main], because they start the runtime first.

src/logging.rs
```diff
@@ -0,0 +1,9 @@
+pub fn init() -> () {
+    let _guard = sentry::init((
+        "https://aa37d7fba870e93bf72c2138b1029d35@o1.ingest.us.sentry.io/4507408549347328",
```
evanpurkhiser, line 3: Will read from a config file once I get to it

wedamija (approved):

PR #43, opened 2024-06-10
feat: Add configuration parsing

evanpurkhiser:


src/logging.rs
```diff
@@ -1,12 +1,38 @@
...
+}
+
+impl From<Config> for LoggingConfig {
+    fn from(config: Config) -> Self {
+        Self {
+            sentry_dsn: config.sentry_dsn,
+            sentry_env: config.sentry_env,
+        }
+    }
+}
```
evanpurkhiser, line 22: Not sure if I TOTALLY love this, but seems nice

src/config.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use figment::{
+    providers::{Env, Format, Yaml},
+    Figment,
+};
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Config {
```
wedamija, line 11: Should this be `SentrySDKConfig`?

evanpurkhiser: Nah this is the global config. I'll add more to it dw

evanpurkhiser: have another PR after this to add the kafka clusters config

src/logging.rs
```diff
@@ -1,12 +1,40 @@
...
```
loewenheim, line 24: I believe this is a misuse of `From`. `A: From<B>` is for cases where `B` can be converted to `A` faithfully and losslessly, which is not the case here (I mean, it's the case now, but it won't be once `Config` gets bigger). You have to throw parts of `Config` away to make it fit into a `LoggingConfig`. See https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from for what I mean.

A better way to do this would be to give `Config` a field `logging: LoggingConfig`.

src/logging.rs
```diff
@@ -1,12 +1,40 @@
...
             release: sentry::release_name!(),
+            environment: config.sentry_env.to_owned(),
             ..Default::default()
-        },
-    ));
+        });
 
-    // We'll release the sentry hub in main.rs
-    std::mem::forget(guard)
+        // We manually deinitalize sentry alter
```
loewenheim, line 37:
```suggestion
        // We manually deinitalize sentry later
```

src/config.rs
```diff
@@ -0,0 +1,69 @@
...
+
+impl Config {
+    /// Load configuration from an optional configuration file and environment
+    pub fn extract(path: &Option<PathBuf>) -> anyhow::Result<Config> {
+        let builder = Figment::new();
+
+        let builder = match path {
+            Some(path) => builder.merge(Yaml::file(path)),
+            _ => builder,
+        };
```
loewenheim, line 27:
```suggestion
        let mut builder = Figment::new();

        if let Some(path) = path {
            builder = builder.merge(Yaml::file(path));
        }
```

src/config.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// The sentry DSN to use for error reporting.
+    pub sentry_dsn: Option<String>,
+
+    /// The environment to report to sentry errors to.
+    pub sentry_env: Option<Cow<'static, str>>,
+}
+
+impl Config {
+    /// Load configuration from an optional configuration file and environment
+    pub fn extract(path: &Option<PathBuf>) -> anyhow::Result<Config> {
```
loewenheim, line 21: `extract` is an odd name for this (I assume it's from the `figment` method?), I would use `load` or somesuch here.

src/config.rs
```diff
@@ -0,0 +1,69 @@
...
+        Jail::expect_with(|jail| {
+            jail.create_file(
+                "config.yaml",
+                r#"
+                sentry_dsn: my_dsn
+                sentry_env: my_env
+                "#,
+            )?;
+
+            jail.set_env("UPTIME_CHECKER_SENTRY_ENV", "my_env_override");
```
loewenheim, line 54: Huh, this is pretty nice.

src/logging.rs
evanpurkhiser: > can be converted to A faithfully and losslessly

Ah I didn't know this ok.

src/config.rs
evanpurkhiser: Yeah i had originally done thus but didn't like that it was mutable

src/config.rs
evanpurkhiser: I'm thinking about later adding a second parameter passing the CliApp so it would extract config from both a file, or from some cli app options.

I'll leave it as extract now, I think it's ok

src/logging.rs
evanpurkhiser: Okay, switched this over to a method on the `LoggingConfig` called `from_config`

loewenheim: You could also make `logging::init` just take two individual parameters, unless you expect `LoggingConfig` to grow

evanpurkhiser: ah yeah.maybe that's easier lol. I'll adjust later I think

wedamija (approved):

PR #46, opened 2024-06-11
feat: Add logging using tracing

evanpurkhiser:


src/logging.rs
```diff
@@ -37,4 +92,31 @@ pub fn init(config: LoggingConfig) {
...
+            .with_current_span(true)
+            .with_span_list(true)
+            .with_file(true)
+            .with_line_number(true)
+            .boxed(),
+    };
+
+    let logs_subscriber = tracing_subscriber::registry()
+        .with(format.with_filter(config.log_level.level_filter()))
+        .with(sentry::integrations::tracing::layer());
```
evanpurkhiser, line 119: most of this stuff is copy / paste from relay

src/logging.rs
```diff
@@ -1,8 +1,55 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Deserialize, Serialize, clap::ValueEnum)]
+#[serde(rename_all = "lowercase")]
+pub enum Level {
+    Error,
+    Warn,
+    Info,
+    Debug,
+    Trace,
+    Off,
+}
```
evanpurkhiser, line 38: Not totally in love with the fact that both of these are deriving `clap::ValueEnum` but oh well

loewenheim: I would suggest using the `tracing` enum `LevelFilter` instead of rolling your own: https://docs.rs/tracing/latest/tracing/level_filters/struct.LevelFilter.html.

evanpurkhiser: Unfortunately I need to derive `clap::ValueEnum`

loewenheim: Ah, I see. Type class coherence/the orphan rule continues to be a massive pain.

loewenheim (approved):

src/logging.rs
```diff
@@ -1,8 +1,55 @@
...
+        match self {
+            Self::Error => LevelFilter::ERROR,
+            Self::Warn => LevelFilter::WARN,
+            Self::Info => LevelFilter::INFO,
+            Self::Debug => LevelFilter::DEBUG,
+            Self::Trace => LevelFilter::TRACE,
+            Self::Off => LevelFilter::OFF,
+        }
+    }
+}
```
loewenheim, line 52: This could also be a `From`/`Into` impl.

evanpurkhiser: I'll switch it over, copied from relay

PR #56, opened 2024-06-14
feat(uptime): Update `sentry-kafka-schemas`, remove monitor ids and add subscription_id

wedamija:
This updates the checker to use `subscription_id` instead of monitor ids

### DIFF
--- a/src/checker.rs
+++ b/src/checker.rs
@@ -123,8 +123,8 @@ impl Checker {
 
         CheckResult {
             guid: Uuid::new_v4(),
-            monitor_id: 0,
-            monitor_environment_id: 0,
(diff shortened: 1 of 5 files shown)


evanpurkhiser (approved):

src/checker.rs
```diff
@@ -26,6 +26,8 @@ pub struct Checker {
     client: Client,
 }
 
+pub const SUBSCRIPTION_ID: Uuid = uuid!("663399a09e6340a79c3c7a3f26878904");
```
evanpurkhiser, line 29: nice

evanpurkhiser (approved):

PR #61, opened 2024-06-17
feat(tracing): Add sentry-trace header

evanpurkhiser:


wedamija (approved):
Do we need to send other options to disable traces from being generated on the server side, and to just include errors?

evanpurkhiser:
> Do we need to send other options to disable traces from being generated on the server side, and to just include errors?

This is a good question

@AbhiPrasad -- I know the last byte is used to say "don't set any sampling rules" but is there any way to tell the SDK to always sample errors during a trace?

AbhiPrasad:
>  I know the last byte is used to say "don't set any sampling rules" but is there any way to tell the SDK to always sample errors during a trace?

Error sampling rate defaults to 100%, so only explicit user config makes them drop errors or reduce that. There is unfortunately no way for us to programmatically change that

PR #65, opened 2024-06-24
feat: Add basic config types

evanpurkhiser:


src/types/check_config.rs
```diff
@@ -0,0 +1,27 @@
+use uuid::Uuid;
+
+/// Valid intervals between the checks.
+#[derive(Debug, Copy, Clone)]
+pub enum CheckInterval {
+    FiveMinutes = 5,
```
wedamija, line 6: Let's allow 1 minute as well, might be useful for our testing with rebalancing etc

wedamija (approved):

src/types/check_config.rs
evanpurkhiser: 👍

Issue #12, opened 2024-05-23
Implement system to take schedule configurations and execute for those schedules


evanpurkhiser:
This will be what actually executes the checks in #11 

## Requirements

- We want to minimize "stampeding". This is when ALL requests happen all at the same time. In the Crons product we suffer from this as people like to have schedules that run on the top of the hour. For uptime monitoring **we control when the checks are made**. This means we can nicely distribute all checks.

- While checks are distributed, check configuration should still run on it's correct interval. If a check is configured to happen every 5 minutes each subsequent check should be 5 minutes apart.Example: if a check is scheduled to run at `2:53:32` the next run must be `2:58:32`

- We can run our checks on some defined tick. For the sake of this ticket, let's say we run them **every second**.

- We should be able to determine which checks need to be run at a given tick in `O(1)` time

- Updates to configurations that change the schedule must be propagated such that the schedule that checks are run is recomputed for that check configuration. 

- We must be able to detect when a check skipped runs to create missed check-ins.

## Configuration structure

Implemented in https://github.com/getsentry/uptime-checker/pull/65

```rs
/// Valid intervals between the checks
#[derive(Debug)]
pub enum CheckInterval {
    FiveMinutes = 5,
    TenMinutes = 10,
    TwentyMinutes = 20,
    ThirtyMintues = 30,
    SixtyMinutes = 60,
}

/// The number of seconds in the biggest interval
pub const MAX_CHECK_INTERVAL_SECS: usize = CheckInterval::SixtyMinutes as usize * 60;

/// The CheckConfig represents a configuration for a single check.
#[derive(Debug)]
pub struct CheckConfig {
    /// The subscription this check configuration is associated to in sentry.
    subscription_id: Uuid,

    /// The interval between each check run.
    interval: CheckInterval,

    /// The actual HTTP URL to check.
    url: String,
}
```

### Assumptions

- We have a maximum interval for checks. For this example, we can say the maximum interval is **1 hour**.
- All smaller intervals are evenly divisible into the maximum interval.

### Approach

We divide our interval into ticks. Since our ticks are on the second boundary this would give us 1 hour worth of slots before the schedule repeats. This is why it is important that all intervals are divisible into the maximum interval, otherwise the pattern does not repeat.

What this might Look like is this

```rs


impl CheckConfig {
    /// Compute at which seconds within the TickBuckets the CheckConfig belongs to. This is
    /// consistent on the subscription_id.
    fn slots(&self) -> Vec<usize> {
[...]

PR #66, opened 2024-06-24
ref(config): Add `slots` fn to determine tick slots

evanpurkhiser:
Part of https://github.com/getsentry/uptime-checker/issues/12

### DIFF
--- a/src/types/check_config.rs
+++ b/src/types/check_config.rs
@@ -1,28 +1,117 @@
 use uuid::Uuid;
 
-/// Valid intervals between the checks.
-#[derive(Debug, Copy, Clone)]
+const ONE_MINUTE: isize = 60;
+
+/// Valid intervals between the checks
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum CheckInterval {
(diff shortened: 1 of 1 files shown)


src/types/check_config.rs
```diff
@@ -1,28 +1,104 @@
...
+/// Valid intervals between the checks
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum CheckInterval {
-    OneMinute = 1,
-    FiveMinutes = 5,
-    TenMinutes = 10,
-    TwentyMinutes = 20,
-    ThirtyMintues = 30,
-    SixtyMinutes = 60,
+    OneMinute = ONE_MINUTE,
```
wedamija, line 8: `Duration.from_mins(1).as_secs()`?

wedamija (approved):
This looks good to me, but I think the tests should confirm the jitter works to put subscriptions into different buckets even if they're on the same schedule

src/types/check_config.rs
evanpurkhiser: Doesn't work at compile time

PR #71, opened 2024-06-25
ref: Use serde_with to serialize times / durations

evanpurkhiser:


markstory (approved):
Looks good to me. serde_with gives some nice ergonomics on deserialization.

PR #80, opened 2024-06-27
feat(check-config): Support msgpack deseraizlization

evanpurkhiser:


src/types/check_config.rs
```diff
@@ -1,10 +1,14 @@
...
+use serde_repr::Deserialize_repr;
+use serde_with::serde_as;
 use uuid::Uuid;
 
 const ONE_MINUTE: isize = 60;
 
 /// Valid intervals between the checks
-#[derive(Debug, Copy, Clone, PartialEq)]
+#[repr(isize)]
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize_repr)]
```
evanpurkhiser, line 11: Needed the `serde_repr` crate to deserialize from `300`. Otherwise, it represents these as `1`, `2`, `3` etc

vartec (approved):

PR #81, opened 2024-06-27
ref(config-store): Prefer Vec for buckets

evanpurkhiser:
These buckets will be fairly small so iteration is fine here


evanpurkhiser:
Going to merge this into a rework that makes the config store compatible with partitons

PR #84, opened 2024-06-28
ref(config-store): Track partitions of configs

evanpurkhiser:
Since we'll be using kafka to store and receive configurations, each config will be part of a kafka partiton. These can be added or removed from the checker during rebalancing, thus we need to support dropping a set of configurations all at once when a partition is removed.

The ConfigStore has learned `drop_partition`, which drops all CheckConfigs associated to a particular partition. It does this efficiently by keeping a TickBuckets list for each partition. When a partition is dropped the entire TickBuckets vector is dropped and all CheckConfigs contained in those buckets are dropped

### DIFF
--- a/src/config_store.rs
+++ b/src/config_store.rs
@@ -1,23 +1,20 @@
-use std::{
-    collections::{HashMap, HashSet},
-    fmt,
-    sync::Arc,
-};
+use std::{collections::HashMap, fmt, sync::Arc};
 
 use chrono::{DateTime, Utc};
 use uuid::Uuid;
 
 use crate::types::check_config::{CheckConfig, MAX_CHECK_INTERVAL_SECS};
 
-/// Maps the unique subscription_id of a CheckConfig to the configuration itself. This provides
-/// O(1) access during configuration updates.
-pub type CheckerConfigs = HashMap<Uuid, Arc<CheckConfig>>;
+// Represents a bucket of checks at a given tick.
+pub type TickBucket = Vec<Arc<CheckConfig>>;
 
-/// The TickBucket is used to determine which checks should be executed at which ticks along the
+/// TickBuckets are used to determine which checks should be executed at which ticks along the
 /// interval pattern. Each tick contains the set of checks that are assigned to that second.
-pub type TickBuckets = Vec<HashSet<Uuid>>;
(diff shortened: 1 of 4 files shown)


davidenwang (approved):

src/config_store.rs
```diff
@@ -82,26 +88,50 @@ impl ConfigStore {
...
-            }
+        let Some(config) = self.configs.remove(&subscription_id) else {
+            return;
+        };
+        let Some(buckets) = self.partitioned_buckets.get_mut(&config.partition) else {
+            return;
+        };
+
+        for slot in config.slots() {
+            buckets[slot].retain(|c| c.subscription_id != subscription_id);
```
wedamija, line 112: Isn't using retain here inefficient? Should we have a data structure that allows o(1) lookups?

src/config_store.rs
```diff
@@ -60,20 +59,27 @@ impl fmt::Display for Tick {
...
-    buckets: TickBuckets,
-    configs: CheckerConfigs,
+    /// A ConfigStore is derived from configurations loaded from Kafka, when the assigned
+    /// partitions for the uptime-checker are re-balanced, partitions will be added and removed.
+    /// Storing TickBuckets for each partition makes dropping an entire partition of configs during
+    /// removal simple.
+    partitioned_buckets: HashMap<i32, TickBuckets>,
+
+    /// Mapping of each subscription_id's config
+    configs: HashMap<Uuid, Arc<CheckConfig>>,
```
wedamija, line 76: I feel like the configs shouldn't be global, but should just be isolated inside each partition, that way they can't influence each other

src/config_store.rs
```diff
@@ -82,26 +88,50 @@ impl ConfigStore {
...
+        };
+
+        buckets
             .iter()
-            .filter_map(|id| self.configs.get(id).cloned())
+            .flat_map(|b| b.iter().cloned())
+            .for_each(|config| {
+                self.configs.remove(&config.subscription_id);
+            });
+    }
```
wedamija, line 128: This is one reason I'd like to not have the configs be global - ideally when we drop a partition we don't need to do anything other than just remove it from some global mapping.

PR #92, opened 2024-07-02
feat: Make ConfigStore aware of known partitions

evanpurkhiser:
This adds a `update_partitions` method the allows the config store to be told what partitions we are responsible for. This method will automatically drop partitions that we become no longer responsible for.

### DIFF
--- a/src/config_store.rs
+++ b/src/config_store.rs
@@ -1,7 +1,9 @@
+use std::collections::HashSet;
 use std::sync::RwLock;
 use std::{collections::HashMap, fmt, sync::Arc};
 
 use chrono::{DateTime, Utc};
+use tracing::{error, trace};
 use uuid::Uuid;
 
 use crate::types::check_config::{CheckConfig, MAX_CHECK_INTERVAL_SECS};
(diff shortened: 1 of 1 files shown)


davidenwang (approved):

davidenwang (commented):
overall makes sense to me, but is there a reason we have to keep track of known_partitions separately? rather than relying on the keys of partitioned_buckets?

evanpurkhiser:
> overall makes sense to me, but is there a reason we have to keep track of known_partitions separately? rather than relying on the keys of partitioned_buckets?

Good point. That does seem better.

davidenwang (approved):
nice nice

src/config_store.rs
```diff
@@ -121,8 +134,39 @@ impl ConfigStore {
...
+        for new_partiton in new_partitions.difference(&known_partitions) {
+            self.partitioned_buckets
+                .entry(*new_partiton)
+                .or_insert_with(make_empty_tick_buckets);
+        }
+    }
+
+    /// Drop an entire partition of check configurations. This does NOT update the known set of
+    /// partitions. Use update_partitions.
+    fn drop_partition(&mut self, partition: u16) {
```
wedamija, line 169: Should this be private so that no one external can call it?

evanpurkhiser: It is, you need `pub` to make it public.

PR #93, opened 2024-07-02
feat: Add config consumer

evanpurkhiser:
This consumer reads config messages from the uptime-configs topic and
registers them in the ConfigStore

### DIFF
--- a/src/config_consumer.rs
+++ b/src/config_consumer.rs
@@ -1,0 +1,238 @@
+use std::{collections::HashMap, sync::Arc, thread};
+
+use chrono::TimeDelta;
+use rust_arroyo::{
+    backends::kafka::{config::KafkaConfig, types::KafkaPayload, InitialOffset},
+    processing::{
+        strategies::{
+            noop::Noop, run_task::RunTask, InvalidMessage, ProcessingStrategy,
+            ProcessingStrategyFactory,
+        },
+        StreamProcessor,
+    },
+    types::{InnerMessage, Message, Partition, Topic},
+};
+use tokio::task::JoinHandle;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info};
+use uuid::Uuid;
+
+use crate::{app::config::Config, config_store::RwConfigStore, types::check_config::CheckConfig};
+
+fn update_config_store(
+    config_store: Arc<RwConfigStore>,
+    message: Message<KafkaPayload>,
+) -> Result<Message<KafkaPayload>, InvalidMessage> {
+    let InnerMessage::BrokerMessage(ref broker_message) = message.inner_message else {
+        panic!("Expected BrokerMessage, got {:?}", message.inner_message);
+    };
+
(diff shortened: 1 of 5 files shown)


src/config_consumer.rs
```diff
@@ -0,0 +1,232 @@
...
+        let _ = update_config_store(config_store.clone(), message);
+
+        let configs = config_store.read().unwrap().all_configs();
+
+        // example_config was removed
+        assert_eq!(configs.len(), 0);
+        assert_eq!(Arc::strong_count(&example_config), 1);
+    }
+
+    // TODO(epurkhiser): We probably want to test the update_partition callback
```
evanpurkhiser, line 231: I think i need to actually run kafka to test this?

src/config_consumer.rs
```diff
@@ -0,0 +1,232 @@
...
+    );
+
+    let mut processing_handle = stream_processor.get_handle();
+
+    let join_handle = thread::spawn(move || {
+        info!("Starting config consumer");
+        stream_processor
+            .run()
+            .expect("Failed to run config consumer");
+    });
```
evanpurkhiser, line 122: Spawn the consumer in it's own thread outside of tokio, no need to treat it as a future.

src/config_consumer.rs
loewenheim: Arroyo contains types for producing/consuming "locally", i.e. without running Kafka (`LocalConsumer`, `LocalBroker`, &c.). You might be able to make it work with those.

loewenheim (approved):

src/config_consumer.rs
```diff
@@ -0,0 +1,242 @@
+use core::panic;
```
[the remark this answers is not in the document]
evanpurkhiser: oops yep

src/config_consumer.rs
```diff
@@ -0,0 +1,242 @@
...
+    };
+
+    let key = broker_message.payload.key().ok_or_else(|| {
+        error!("Missing message key (subscription_id)");
+        InvalidMessage::from(broker_message)
+    })?;
+    let subscription_id = Uuid::from_slice(key).map_err(|err| {
+        error!(
+            ?err,
+            "Key must be a valid subscription_id UUID. Got {:?}", key
```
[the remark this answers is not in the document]
evanpurkhiser: Ah good call out. Thank you. I assume if I don't format it and add it as an earlier parameter it will be OK?

src/config_consumer.rs
evanpurkhiser: cool, I'll try and dig into this.

src/config_consumer.rs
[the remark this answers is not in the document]
loewenheim: Yeah, then it will be added as a tag.

src/config_consumer.rs
evanpurkhiser: Will follow up on this later. Going to merge for now.

PR #97, opened 2024-07-08
ref: Extract scheduler_loop function with trait impl params

evanpurkhiser:
Will allow us to write tests for this code using a mock / noop checker and producer


untitaker (approved):
looks good to me (I don't know what's going on)

davidenwang (approved):

Issue #98, opened 2024-07-08
Instrument uptime checker


evanpurkhiser:
There's a number of things that are going to be important to keep an eye on

- [x] How long does it take for a checker to read the configuration at boot. This can be instrumented by keeping track of how long it takes for us to read a message that is newer than the time the consumer began reading.

- [x] A gauge on how many checks we're scheduling per bucket.

- [x] A gauge on the latency from the time the check was scheduled to when it completed execution

- [x] A gauge on how long checks take to execute. (May not be that useful since different domains will have different timeouts)

- [x] A histogram of success/failure including failure reasons. We'll want to keep an eye that we don't have any huge spikes of problems.

- [ ] If we implement an execution pool with concurrency (we'll probably have to) then we'll want to keep an eye on the backlog for this.

PR #100, opened 2024-07-10
feat(metris): Record a number of metrics

evanpurkhiser:
- Check duration
 - Checks per bucket
 - Check status
 - Check failure reason

Part of https://github.com/getsentry/uptime-checker/issues/98

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -14,6 +14,7 @@ use crate::checker::Checker;
 use crate::config_store::{RwConfigStore, Tick};
 use crate::producer::kafka_producer::KafkaResultsProducer;
 use crate::producer::ResultsProducer;
+use crate::types::result::{CheckResult, CheckStatus, CheckStatusReasonType};
 
 pub fn run_scheduler(
     config: &Config,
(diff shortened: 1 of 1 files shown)


src/scheduler.rs
```diff
@@ -30,6 +32,54 @@ pub fn run_scheduler(
...
+    } = result;
+
+    // Record duration of check
+    if let Some(duration) = duration {
+        Metric::timing("http_checker.duration", duration.to_std().unwrap()).send();
+    }
+
+    // Record time between scheduled and actual check
+    Metric::timing(
+        "http_checker.duration",
```
wedamija, line 53: Need to change this name? `drift` or something?

src/scheduler.rs
```diff
@@ -30,6 +32,54 @@ pub fn run_scheduler(
...
+        scheduled_check_time,
+        actual_check_time,
+        duration,
+        status_reason,
+        ..
+    } = result;
+
+    // Record duration of check
+    if let Some(duration) = duration {
+        Metric::timing("http_checker.duration", duration.to_std().unwrap()).send();
```
wedamija, line 48: Should we have tags with the result on these (and all metrics)?

wedamija (approved):

evanpurkhiser:
nvm we're not using this, we want datadog

src/scheduler.rs
evanpurkhiser: good idea 👍

src/scheduler.rs
evanpurkhiser: Changed to `delay`

src/scheduler.rs
```diff
@@ -30,6 +31,63 @@ pub fn run_scheduler(
...
+        "check_result.status",
+        "status" => status_label,
+        "failure_reason" => failure_reason.unwrap_or("ok"),
+    )
+    .increment(1);
+
+    // Record failure reason
+    if let Some(failure_reason) = failure_reason {
+        metrics::counter!("check_result.failure_reason", "reason" => failure_reason).increment(1);
+    }
```
wedamija, line 88: We don't really need this do we? Isn't it inherently included in the `status` counter?

src/scheduler.rs
```diff
@@ -30,6 +31,63 @@ pub fn run_scheduler(
...
+        "check_result.delay_ms",
+        "histogram" => "timer",
+        "status" => status_label,
+        "failure_reason" => failure_reason.unwrap_or("ok"),
+    )
+    .record((*actual_check_time - *scheduled_check_time).num_milliseconds() as f64);
+
+    // Record status of the check
+    metrics::counter!(
+        "check_result.status",
```
wedamija, line 79: I think maybe this should be `check_result.processed` or something? The status is already a tag

src/scheduler.rs
evanpurkhiser: hmm, this one only logs when there's a real failure. I guess we should be able to exclude the `none` and have a failure only chart in dd?

PR #105, opened 2024-07-16
ref(build): Remove cargo update steps

evanpurkhiser:
This was updating packages. We don't actually want that


wedamija (approved):
I feel like I needed this so that things would actually build at all... I guess it's ok now

PR #108, opened 2024-07-22
Refactor service per partition

evanpurkhiser:
Refactors the uptime checker to have a separate scheduler per Kafka partition. This makes assigning/revoking partitions much simpler, and allows us to more easily block while partitions are loading in the future.


wedamija (approved):
lgtm, I did extra work here to finish this up and will merge these together https://github.com/getsentry/uptime-checker/pull/112

PR #118, opened 2024-07-31
ref: Only start new partitions

evanpurkhiser:


wedamija (approved):

evanpurkhiser:
Changed this implementation a bit if you could take a look @wedamija

wedamija (approved):

PR #120, opened 2024-07-31
fix: Run consumer in it's own thread

evanpurkhiser:
We don't want this running as a tokio task since it will block and consume an entire thread. Tokio will only have so many task executor threads so we may end up the scenario where executors are available to drive futures forward.

The reason it was run in a tokio task before was because the partitions update function in the manager would attempt to start new tokio tasks, which it could not do outside of the tokio runtime.

Instead we can use tokios `task::spawn_blocking` to run the consumer, which tokio will run in it's own blocking thread.

### DIFF
--- a/src/config_consumer.rs
+++ b/src/config_consumer.rs
@@ -11,7 +11,7 @@ use rust_arroyo::{
     types::{InnerMessage, Message, Partition, Topic},
 };
 use std::{collections::HashMap, sync::Arc};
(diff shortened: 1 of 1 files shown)


wedamija (approved):
So the consumer runs in a thread, and the update_partitions call runs in tokio in a task?

Seems weird that the consumer is unable to work with tokio to begin with

wedamija (approved):

PR #122, opened 2024-08-01
feat: Add metrics for each partitions "boot"

evanpurkhiser:
Brings back the logic we wrote to detect when a partition has "completed" booting

### DIFF
--- a/src/config_waiter.rs
+++ b/src/config_waiter.rs
@@ -1,3 +1,12 @@
+use std::{sync::Arc, time::Duration};
+
+use tokio::{
+    sync::oneshot::{self, Receiver},
+    time::{interval, Instant},
+};
+
+use crate::config_store::RwConfigStore;
+
 /// How long does the consumer need to be idle before we consider it to have "finished" reading the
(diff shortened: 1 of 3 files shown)


src/config_waiter.rs
```diff
@@ -55,23 +64,37 @@ fn wait_for_boot(config_store: Arc<RwConfigStore>) -> Receiver<()> {
...
-        metrics::gauge!("config_consumer.total_configs").set(total_configs as f64);
+        tracing::info!(
+            boot_time_ms,
+            total_configs,
+            partition,
+            "partition_boot_complete"
+        );
+
+        metrics::gauge!("config_consumer.partition_boot_time_ms").set(boot_time_ms as f64);
+        metrics::gauge!("config_consumer.partition_total_configs").set(total_configs as f64);
```
wedamija, line 75: Should these have partition as a tag?

evanpurkhiser: Yeah let me add it 👍

wedamija (approved):

PR #124, opened 2024-08-01
ref: Include partition ID in scheduler startup

evanpurkhiser:


wedamija (approved):

src/scheduler.rs
```diff
@@ -12,12 +12,13 @@ use crate::producer::ResultsProducer;
...
 
 pub fn run_scheduler(
+    partition: u16,
     config_store: Arc<RwConfigStore>,
     checker: Arc<impl Checker + 'static>,
     producer: Arc<impl ResultsProducer + 'static>,
     shutdown: CancellationToken,
 ) -> JoinHandle<()> {
-    info!("Starting scheduler");
+    info!("Starting scheduler on partition {}", partition);
```
evanpurkhiser, line 21: Actually let me switch to using a parameter

PR #125, opened 2024-08-01
chore(mocks): Add mock_checker and mock_producer

wedamija:
Add in mock checker and producer for use in testing

### DIFF
--- a/src/checker/mock_checker.rs
+++ b/src/checker/mock_checker.rs
@@ -1,0 +1,43 @@
+use chrono::Utc;
+use sentry::protocol::{SpanId, TraceId};
+use uuid::Uuid;
+use crate::checker::Checker;
+use crate::config_store::Tick;
+use crate::types::check_config::CheckConfig;
+use crate::types::result::{CheckResult, CheckStatus};
+
+#[derive(Clone, Debug)]
+pub struct MockChecker {}
+impl MockChecker {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+impl Checker for MockChecker {
+    /// Makes a request to a url to determine whether it is up.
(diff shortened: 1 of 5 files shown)


src/checker/mock_checker.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct MockChecker {}
+impl MockChecker {
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+impl Checker for MockChecker {
+    /// Makes a request to a url to determine whether it is up.
+    /// Up is defined as returning a 2xx within a specific timeframe.
+    #[tracing::instrument]
```
evanpurkhiser, line 19: Can drop the instrument here I thnk

src/producer/mock_producer.rs
```diff
@@ -0,0 +1,34 @@
...
+    topic: TopicOrPartition,
+    schema: Schema,
+}
+
+impl MockResultsProducer {
+    pub fn new(topic_name: &str, config: KafkaConfig) -> Self {
+        let producer = KafkaProducer::new(config);
+        let topic = TopicOrPartition::Topic(Topic::new(topic_name));
+        let schema = sentry_kafka_schemas::get_schema("uptime-results", None).unwrap();
+
```
evanpurkhiser, line 19: Instead of this actually using kafka for testing, hy don't we have it use a channel, and we can just make sure something got put in the channel in the tests

evanpurkhiser (approved):

src/producer/mock_producer.rs
wedamija: Oh this doesn't use anything for testing, it just uses the kafka stuff to get the schema and validate

PR #127, opened 2024-08-01
feat(uptime): Store completed ticks in redis

wedamija:
This stores tick progress in Redis and loads it when the scheduler starts up. We use this to set the instance that we start from, so that we won't miss ticks during deploys, rebalances and crashes.

We store the ticks as nanos and use them to reset the Interval back in time. We store the completed ticks in `in_order_tick_processor_join_handle`, which gives us a guaranteed in-order completion of ticks, so that we know we're not missing anything.
@wedamija

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - uses: shogo82148/actions-setup-redis@v1
+      - uses: shogo82148/actions-setup-redis@428a144f11914412a8f52902296e134dd2e0f925
         with:
           redis-version: "7.x"
       - name: Install Rust Toolchain
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - uses: shogo82148/actions-setup-redis@v1
+        with:
+          redis-version: "7.x"
       - name: Install Rust Toolchain
         run: rustup toolchain install stable --profile minimal --no-self-update
       - name: Run Cargo Tests
         run: cargo test
+
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -56,9 +56,8 @@ pub struct Config {
     /// The topic to load [`CheckConfig`]s from.
     pub configs_kafka_topic: String,
 
-    /// The general purpose redis cluster to use with this service
-    #[serde_as(as = "serde_with::StringWithSeparator::<CommaSeparator, String>")]
-    pub redis_cluster_nodes: Vec<String>,
+    /// The general purpose redis node to use with this service
+    pub redis_host: String,
 }
 
 impl Default for Config {
@@ -74,7 +73,7 @@ impl Default for Config {
             results_kafka_topic: "uptime-results".to_owned(),
             configs_kafka_cluster: vec![],
             configs_kafka_topic: "uptime-configs".to_owned(),
-            redis_cluster_nodes: vec![],
+            redis_host: "redis://127.0.0.1:6379".to_owned(),
         }
     }
 }
@@ -127,7 +126,7 @@ mod tests {
                 results_kafka_cluster: '10.0.0.1,10.0.0.2:9000'
                 configs_kafka_cluster: '10.0.0.1,10.0.0.2:9000'
                 statsd_addr: 10.0.0.1:8126
-                redis_cluster_nodes: '10.0.0.1:6379,10.0.0.2:6379'
+                redis_host: redis://127.0.0.1:6379
                 "#,
             )?;
 
@@ -153,10 +152,7 @@ mod tests {
                     configs_kafka_cluster: vec!["10.0.0.1".to_owned(), "10.0.0.2:9000".to_owned()],
                     results_kafka_topic: "uptime-results".to_owned(),
                     configs_kafka_topic: "uptime-configs".to_owned(),
-                    redis_cluster_nodes: vec![
-                        "10.0.0.1:6379".to_owned(),
-                        "10.0.0.2:6379".to_owned()
-                    ],
+                    redis_host: "redis://127.0.0.1:6379".to_owned(),
                 }
             );
             Ok(())
@@ -185,10 +181,7 @@ mod tests {
                 "10.0.0.1,10.0.0.2:7000",
             );
             jail.set_env("UPTIME_CHECKER_STATSD_ADDR", "10.0.0.1:1234");
-            jail.set_env(
-                "UPTIME_CHECKER_REDIS_CLUSTER_NODES",
-                "10.0.0.3:6379,10.0.0.2:6379",
-            );
+            jail.set_env("UPTIME_CHECKER_REDIS_HOST", "10.0.0.3:6379");
             let app = cli::CliApp {
                 config: Some(PathBuf::from("config.yaml")),
                 log_level: Some(logging::Level::Trace),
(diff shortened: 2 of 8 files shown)


src/scheduler.rs
```diff
@@ -140,6 +155,50 @@ async fn scheduler_loop(
 
         debug!(tick = %tick, "Scheduler ticking");
         schedule_checks(tick);
+        let _: () = connection.set(&progress_key, tick.time().timestamp_nanos_opt().unwrap()).await.expect("Couldn't save progress of scheduler");
```
wedamija, line 158: Realised that setting started above means we'll start behind, but never actually catch up since we're always waiting for interval between ticks. Will update to either not overwrite start for now, or to have it perform a quick catch up whenever the current tick to process is < 1 tick behind.

.github/workflows/test.yml
```diff
@@ -9,6 +9,17 @@ jobs:
     timeout-minutes: 5
     runs-on: ubuntu-latest
     steps:
+      - name: Test redis cluster
+        uses: vishnudxb/redis-cluster@1.0.9
```
wedamija, line 13: Is it ok to use some external action like this?

evanpurkhiser: Does this mean these tests won't run at all without redis locally either? @jan-auer wondering how we might mark acceptance tests like this different from our typical unit tests.

evanpurkhiser: @asottile-sentry We usually pin versions by hash right?

src/app/config.rs
```diff
@@ -70,7 +70,9 @@ impl Default for Config {
             results_kafka_topic: "uptime-results".to_owned(),
             configs_kafka_cluster: vec![],
             configs_kafka_topic: "uptime-configs".to_owned(),
-            redis_cluster_nodes: vec![],
+            redis_cluster_nodes: (0..6)
+                .map(|i| format!("redis://127.0.0.1:700{}", i))
+                .collect::<Vec<_>>(),
```
evanpurkhiser, line 75: Is this the normal default?

wedamija: Right, we should add a README at some point. You can run a redis cluster in sentry by setting `SENTRY_DEV_USE_REDIS_CLUSTER = True`. This ip address points to that cluster

.github/workflows/test.yml
wedamija: > Does this mean these tests won't run at all without redis locally either? @jan-auer wondering how we might mark acceptance tests like this different from our typical unit tests.

Yeah we'll require redis, you can run it using devservices. For now that seems simpler then making our own devservices

evanpurkhiser: Yeah agree. Just wondering if there should be a way to mark these as like "acceptance" tests

src/scheduler.rs
```diff
@@ -78,10 +92,29 @@ async fn scheduler_loop(
...
+    let result: Option<String> = connection
+        .get(&progress_key)
+        .await
+        .expect("Unable to get last tick key");
+    let start = match result {
+        Some(result) => {
+            let nanos: i64 = result
+                .parse()
+                .expect("Last tick timestamp has invalid format");
+            Utc.timestamp_nanos(nanos)
```
evanpurkhiser, line 112: interesting, Why isn't this in the form `Utc::timestamp_nanos`

src/scheduler.rs
```diff
@@ -140,6 +173,66 @@ async fn scheduler_loop(
 
         debug!(tick = %tick, "Scheduler ticking");
         schedule_checks(tick);
+        let _: () = connection
```
evanpurkhiser, line 176: Why did you have to type this?
```suggestion
        let _ = connection
```

wedamija: Rust complains otherwise 🤷

src/scheduler.rs
```diff
@@ -78,10 +92,29 @@ async fn scheduler_loop(
...
     shutdown: CancellationToken,
+    progress_key: String,
+    redis_cluster_nodes: Vec<String>,
 ) {
-    let mut interval = time::interval(time::Duration::from_secs(1));
+    let client = ClusterClient::new(redis_cluster_nodes).unwrap();
+    let mut connection = client
+        .get_async_connection()
+        .await
+        .expect("Unable to connect to Redis");
```
evanpurkhiser, line 102: Is it OK to do this every time? would it make more sense to pass the connection in when we start the scheduler?

evanpurkhiser: oh nvm I see. ignore me

src/scheduler.rs
wedamija: Actually also mentioned in the docs for redis-rs https://github.com/redis-rs/redis-rs

evanpurkhiser (approved):

.github/workflows/test.yml
asottile-sentry: yeah it would be best to pin this action to a hash which is good enough to prevent drift

we'll hopefully have some stuff *soon* which makes this easier (without needing to use `sentry devservices ...`) and will allow any service at sentry to have reusable dependent services without needing `sentry` at all (basically `devservices` but independent and better!)

src/checker.rs
```diff
@@ -1,3 +1,4 @@
+pub(crate) mod dummy_checker;
```
evanpurkhiser, line 1: ah smart

src/scheduler.rs
```diff
@@ -78,11 +92,28 @@ async fn scheduler_loop(
...
+    let result: Option<String> = connection
+        .get(&progress_key)
+        .await
+        .expect("Unable to get last tick key");
+    let tick_frequency = Duration::from_secs(1);
+    let mut current_tick_time = match result {
+        Some(result) => {
+            let nanos: i64 = result
+                .parse()
+                .expect("Last tick timestamp has invalid format");
```
evanpurkhiser, line 112: This will crash the scheduler? Do we want that or should we just return `Utc::now`

src/producer/dummy_producer.rs
```diff
@@ -8,7 +8,7 @@ pub struct DummyResultsProducer {
 
 impl DummyResultsProducer {
     pub fn new(_topic_name: &str) -> Self {
-        let schema = sentry_kafka_schemas::get_schema("uptime-results", None).unwrap();
+        let schema = sentry_kafka_schemas::get_schema(_topic_name, None).unwrap();
```
evanpurkhiser, line 11: Let's remove the `_` prefix

src/scheduler.rs
```diff
@@ -1,33 +1,71 @@
...
+        .await
+        .expect("Unable to get last tick key");
+    let tick_frequency = Duration::from_secs(1);
+    tracing::debug!(progress_key, result, "scheduler.redis_stored_tick_value");
+    let start = match result {
+        Some(result) => {
+            let nanos: i64 = result.parse().unwrap_or(Utc::now().timestamp());
+            Utc.timestamp_nanos(nanos) + tick_frequency
+        }
+        None => Utc::now().duration_trunc(TimeDelta::seconds(1)).unwrap(),
```
evanpurkhiser, line 60: Probably worth a comment here

src/scheduler.rs
```diff
@@ -1,33 +1,71 @@
...
+    let result: Option<String> = connection
+        .get(&progress_key)
+        .await
+        .expect("Unable to get last tick key");
+    let tick_frequency = Duration::from_secs(1);
+    tracing::debug!(progress_key, result, "scheduler.redis_stored_tick_value");
+    let start = match result {
+        Some(result) => {
+            let nanos: i64 = result.parse().unwrap_or(Utc::now().timestamp());
+            Utc.timestamp_nanos(nanos) + tick_frequency
```
evanpurkhiser, line 58: We add one since this was the most recent completed check right. maybe a better name would be

```suggestion
            let last_completed_check_nanos: i64 = result.parse().unwrap_or(Utc::now().timestamp());
            Utc.timestamp_nanos(last_completed_check_nanos) + tick_frequency
```

src/scheduler.rs
```diff
@@ -1,33 +1,71 @@
...
+        }
+        None => Utc::now().duration_trunc(TimeDelta::seconds(1)).unwrap(),
+    };
+    tracing::info!(%start, "scheduler.starting_at");
 
-    let start = Utc::now();
-    let instant = Instant::now();
+    let instant = Instant::now()
+        .checked_sub((Utc::now() - start).to_std().unwrap())
+        .unwrap();
```
evanpurkhiser, line 66: Does this align it to the second then right

evanpurkhiser (approved):

src/scheduler.rs
wedamija: We align to the second here: `Utc::now().duration_trunc(TimeDelta::seconds(1)).unwrap()`

We could also just always align to the second if we want to be safe

PR #129, opened 2024-08-02
ref: Prefix logging calls with `tracing::`

evanpurkhiser:
Just a little bit of consistency, though no way to enforce this so 🤷‍♂️


evanpurkhiser:
Replacing with https://github.com/getsentry/uptime-checker/pull/131

PR #132, opened 2024-08-02
feat(executor): Implement separate executor

evanpurkhiser:


src/app/config.rs
```diff
@@ -63,6 +66,7 @@ impl Default for Config {
         Self {
             sentry_dsn: None,
             sentry_env: None,
+            checker_concurrency: 200,
```
evanpurkhiser, line 69: Not sure if 200 is a good default here

PR #135, opened 2024-08-02
ref: Avoid stop / start state in Manager / PartitionedService

evanpurkhiser:
Instead of the manager having "stopped" and "started" state, this makes it so you may ONLY create a manager that is started.

Same for the PartitionedService.

This also makes it easier in the future now to wait for partitioned services to fully stop before the finishing manager shutdown.

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -6,6 +6,7 @@ use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, RwLock},
 };
+use tokio::task::JoinHandle;
 use tokio_util::sync::CancellationToken;
 
 use crate::config_waiter::wait_for_partition_boot;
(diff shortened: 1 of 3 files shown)


src/manager.rs
```diff
@@ -28,90 +29,89 @@ pub struct PartitionedService {
...
-    pub fn new(config: Arc<Config>) -> Self {
-        Self {
-            config,
-            services: RwLock::new(HashMap::new()),
-            shutdown_signal: CancellationToken::new(),
-        }
-    }
-
-    /// Instantiates a new manager with a single partition and a default config
-    pub fn new_simple() -> Self {
```
evanpurkhiser, line 97: Removed this helper. Felt weird to have a test helper like this in the impl anyway.

src/manager.rs
```diff
@@ -28,90 +29,89 @@ pub struct PartitionedService {
...
 impl PartitionedService {
-    pub fn new(config: Arc<Config>, partition: u16) -> Self {
-        Self {
-            config,
-            partition,
-            config_store: Arc::new(ConfigStore::new_rw()),
-            shutdown_signal: CancellationToken::new(),
-        }
-    }
+    pub fn start(config: Arc<Config>, partition: u16) -> (Self, JoinHandle<()>) {
```
evanpurkhiser, line 32: Starting the partitioned service now returns the join handle for the scheduler.

We don't wait for this yet, but in a future commit we can.

src/manager.rs
```diff
@@ -28,90 +29,89 @@ pub struct PartitionedService {
...
+        };
+
+        (service, scheduler_join_handle)
     }
 
     pub fn get_partition(&self) -> u16 {
         self.partition
     }
 
+    pub fn get_config_store(&self) -> Arc<RwConfigStore> {
```
evanpurkhiser, line 71: This isn't new, the diff is just kind of bad.

src/manager.rs
```diff
@@ -28,90 +29,89 @@ pub struct PartitionedService {
...
-        self.shutdown_signal.clone()
+
+        let service = Self {
+            config,
+            partition,
+            config_store,
+            shutdown_signal,
+        };
+
+        (service, scheduler_join_handle)
```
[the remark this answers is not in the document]
evanpurkhiser: Yeah I agree with this. Let's merge as is and I'll clean this up right after

wedamija (approved):

PR #136, opened 2024-08-02
feat: Add check_executor

evanpurkhiser:
This adds a new async task that reads from an unbounded mpsc channel and executes checks, firing the oneshot channel when complete.

This will be used in place of the scheduler actually firing off the http checker tasks

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -1,0 +1,243 @@
+use std::sync::Arc;
+
+use futures::StreamExt;
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::oneshot::{self, Receiver, Sender};
+use tokio::task::JoinHandle;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+
+use crate::checker::Checker;
+use crate::config_store::Tick;
+use crate::producer::ResultsProducer;
+use crate::types::check_config::CheckConfig;
+use crate::types::result::{CheckResult, CheckStatus, CheckStatusReasonType};
+
+#[derive(Debug)]
+pub struct ScheduledCheck {
+    tick: Tick,
+    config: Arc<CheckConfig>,
+    resolve_tx: Sender<CheckResult>,
+}
+
+pub type CheckSender = UnboundedSender<ScheduledCheck>;
+
+pub fn run_executor(
+    concurrency: usize,
+    checker: Arc<impl Checker + 'static>,
+    producer: Arc<impl ResultsProducer + 'static>,
+) -> (CheckSender, JoinHandle<()>) {
+    tracing::info!("Starting check executor");
+
+    let (sender, reciever) = mpsc::unbounded_channel();
+    let executor =
+        tokio::spawn(async move { executor_loop(concurrency, checker, producer, reciever).await });
+
+    (sender, executor)
+}
+
+pub fn queue_check(
+    sender: &CheckSender,
+    tick: Tick,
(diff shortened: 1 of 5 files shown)


src/check_executor.rs
```diff
@@ -0,0 +1,169 @@
...
+    use tokio::sync::oneshot;
+    use uuid::Uuid;
+
+    use super::*;
+    use crate::{
+        checker::dummy_checker::DummyChecker, producer::dummy_producer::DummyResultsProducer,
+    };
+
+    #[tokio::test]
+    async fn test_executor_simple() {
```
wedamija, line 143: Worth having a test where the producer errors?

src/check_executor.rs
```diff
@@ -0,0 +1,169 @@
...
+                    tick,
+                    resolve_tx,
+                } = scheduled_check;
+
+                let _ = tokio::spawn(async move {
+                    let check_result = job_checker.check_url(&config, &tick).await;
+
+                    if let Err(e) = job_producer.produce_checker_result(&check_result) {
+                        tracing::error!(error = ?e, "Failed to produce check result");
+                    }
```
wedamija, line 63: If we error here should we skip recording results and so on? I'm actually not sure what's best since the result failed to produce, so we kind of want to retry or block or something, since otherwise we're dropping results.

src/check_executor.rs
```diff
@@ -0,0 +1,169 @@
...
+    reciever: UnboundedReceiver<ScheduledCheck>,
+) {
+    let schedule_check_stream: UnboundedReceiverStream<_> = reciever.into();
+
+    schedule_check_stream
+        .for_each_concurrent(concurrency, |scheduled_check| {
+            let job_checker = checker.clone();
+            let job_producer = producer.clone();
+
+            // TODO(epurkhiser): Record metrics on the size of the size of the
```
wedamija, line 49: Bucket?

wedamija (approved):

evanpurkhiser:
Yeah will get a few more tests in first

src/check_executor.rs
evanpurkhiser: Yeah, though I'm still wondering what it should actually do.

src/check_executor.rs
```diff
@@ -0,0 +1,243 @@
...
+        time::sleep(Duration::from_millis(100)).await;
+        assert_eq!(poll!(resolve_rx.as_mut()), Poll::Pending);
+
+        let result = resolve_rx.await.unwrap();
+        assert_eq!(result.subscription_id, config.subscription_id);
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_executor_concurrent_limit() {
+        let checker = Arc::new(DummyChecker::new(Duration::from_secs(1)));
```
wedamija, line 195: Should we just make this 100ms so that test isn't too slow?

src/check_executor.rs
wedamija: Could be something we handle as a separate task. I think that the produce here isn't actually a guarantee that we wrote to Kafka anyway, there's a callback for that from what I remember

src/check_executor.rs
evanpurkhiser: Time is frozen, so it auto advances

wedamija: Lol right, I keep forgetting this

src/check_executor.rs
evanpurkhiser: Yeah let's handle this separately I think.

PR #137, opened 2024-08-02
feat: Add delay to dummy checker

evanpurkhiser:


src/checker/dummy_checker.rs
```diff
@@ -1,18 +1,26 @@
...
 #[derive(Clone, Debug)]
-pub struct DummyChecker {}
+pub struct DummyChecker {
+    delay: Duration,
+}
+
 impl DummyChecker {
-    pub fn new() -> Self {
-        Self {}
+    pub fn new(delay: Duration) -> Self {
```
wedamija, line 19: Does it make sense to make it optional?

wedamija (approved):

PR #145, opened 2024-08-05
feat: Add task to wait for tick completion in order

evanpurkhiser:
Will add a test for this, but wanted to get it up for eyes first

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -31,10 +31,17 @@ impl ScheduledCheck {
     pub fn get_tick(&self) -> &Tick {
         &self.tick
     }
+
+    /// Report the completion of the scheduled check.
+    pub fn record_result(self, result: CheckResult) {
+        self.resolve_tx
+            .send(result)
+            .expect("Failed to resolve completed check");
+    }
 }
 
 impl CheckResult {
-    /// Produce a missed check result from a scheduld check.
+    /// Produce a missed check result from a scheduled check.
     pub fn missed_from(config: &CheckConfig, tick: &Tick) -> Self {
         Self {
             guid: Uuid::new_v4(),
@@ -67,6 +74,8 @@ pub fn run_executor(
     (sender, executor)
 }
 
+/// Queues a check for execution, returning a oneshot receiver that will be fired once the check
+/// has resolved to a CheckResult.
 pub fn queue_check(
     sender: &CheckSender,
     tick: Tick,
@@ -124,10 +133,7 @@ async fn executor_loop(
                     record_result_metrics(&check_result);
                     tracing::info!(result = ?check_result, "executor.check_complete");
 
-                    scheduled_check
-                        .resolve_tx
-                        .send(check_result)
-                        .expect("Failed to resolve completed check");
+                    scheduled_check.record_result(check_result);
                 })
                 .await;
             }
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1,6 +1,7 @@
 use std::sync::Arc;
 
 use chrono::Utc;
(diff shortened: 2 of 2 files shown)


src/check_executor.rs
```diff
@@ -29,6 +29,13 @@ impl ScheduledCheck {
     pub fn get_tick(&self) -> &Tick {
         &self.tick
     }
+
+    /// Report the completion of the scheulded check.
+    pub fn record_result(self, result: CheckResult) {
```
evanpurkhiser, line 34: Adding this here so we can report the results back in tests

src/check_executor.rs
```diff
@@ -67,6 +74,8 @@ pub fn run_executor(
     (sender, executor)
 }
 
+/// Queues a check for execution, returning a oneSshot reciever that will be fired once the check
```
wedamija, line 77: *oneshot
*receiver

src/scheduler.rs
```diff
@@ -65,32 +68,62 @@ async fn scheduler_loop(
...
+            tick
         });
+
+        tick_complete_tx
+            .send(tick_complete_join_handle)
+            .expect("Failed to queue join handle for tick completion");
     };
 
+    let in_order_tick_processor_join_handle = tokio::spawn(async move {
+        while let Some(tick_complete_join_handle) = tick_complete_rx.recv().await {
```
wedamija, line 80: What case would this be `None`? Just when we terminate things?

Is there any chance it can return `None` if there's nothing to `.recv`?

src/scheduler.rs
```diff
@@ -65,32 +68,62 @@ async fn scheduler_loop(
...
+    });
+
     while !shutdown.is_cancelled() {
         let interval_tick = interval.tick().await;
         let tick = Tick::from_time(start + interval_tick.duration_since(instant));
         schedule_checks(tick);
     }
+
+    // Wait for all dispatch ticks to have completed
+    drop(tick_complete_tx);
```
wedamija, line 98: This basically just closes the channel? Does that end up causing `.recv` to return None so that the loop exits?

wedamija (approved):
lgtm,  just some typos and questions for my own understanding

src/scheduler.rs
```diff
@@ -65,32 +68,62 @@ async fn scheduler_loop(
...
         });
+
+        tick_complete_tx
+            .send(tick_complete_join_handle)
+            .expect("Failed to queue join handle for tick completion");
     };
 
+    let in_order_tick_processor_join_handle = tokio::spawn(async move {
+        while let Some(tick_complete_join_handle) = tick_complete_rx.recv().await {
+            // XXX: This task guarantees that we can process ticks IN ORDER upon tick execution.
```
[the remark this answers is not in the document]
evanpurkhiser: yeah, `XXX` to me is usually like "look here!" or "important!".

Though maybe `NOTE` is better

src/scheduler.rs
evanpurkhiser: When the channel closes it becomes none. otherwise it waits to receive more things

https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv

>The channel is closed when all senders have been dropped, or when [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close) is called.

src/scheduler.rs
evanpurkhiser: yeah exactly

evanpurkhiser: actually I think I can make it more explicit.

evanpurkhiser: nvm this is the best way

PR #147, opened 2024-08-05
feat: Add some debugging

evanpurkhiser:


src/check_executor.rs
```diff
@@ -185,6 +188,11 @@ fn record_result_metrics(result: &CheckResult) {
     )
     .record((*actual_check_time - *scheduled_check_time).num_milliseconds() as f64);
 
+    tracing::info!(
+        "delay is {}",
+        (*actual_check_time - *scheduled_check_time).num_milliseconds() as f64
+    );
+
```
wedamija, line 195: It's duration that is incorrect right?

wedamija (approved):

src/check_executor.rs
evanpurkhiser: oh you're right lol

evanpurkhiser: I was using this to check something else I think

PR #152, opened 2024-08-06
feat: Add user agent for checks

evanpurkhiser:
Matches what we did over here https://github.com/getsentry/sentry/pull/75673

### DIFF
--- a/src/checker/http_checker.rs
+++ b/src/checker/http_checker.rs
@@ -1,4 +1,5 @@
 use chrono::{TimeDelta, Utc};
+use reqwest::header::HeaderMap;
 use reqwest::{Client, ClientBuilder, Response};
 use sentry::protocol::{SpanId, TraceId};
 use std::error::Error;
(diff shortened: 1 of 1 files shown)


src/checker/http_checker.rs
```diff
@@ -63,7 +67,11 @@ fn dns_error(err: &reqwest::Error) -> Option<String> {
 
 impl HttpChecker {
     pub fn new() -> Self {
+        let mut default_headers = HeaderMap::new();
+        default_headers.insert("User-Agent", UPTIME_USER_AGENT.to_string().parse().unwrap());
```
wedamija, line 71: Shouldn't we be using the agent from the config?

evanpurkhiser: we don't have it in the config yet do we?

evanpurkhiser: We can slap it in there if we want

wedamija: Oh right, yeah we can do that whenever then. It's more important for when people are able to configure their own uptime monitors

wedamija (approved):

PR #153, opened 2024-08-08
fix(uptime): Make scheduler wait for boot before it starts scheduling ticks

wedamija:
We need to wait until all configs are loaded before we start scheduling ticks. Otherwise we end up missing configs due to a race condition between loading configs and processing ticks.

### DIFF
--- a/src/config_waiter.rs
+++ b/src/config_waiter.rs
@@ -1,11 +1,11 @@
 use std::{sync::Arc, time::Duration};
 
+use crate::config_store::RwConfigStore;
 use tokio::{
     sync::oneshot::{self, Receiver},
     time::{interval, Instant},
 };
-
-use crate::config_store::RwConfigStore;
+use tokio_util::sync::CancellationToken;
 
 /// How long does the consumer need to be idle before we consider it to have "finished" reading the
 /// backlog of configs.
@@ -30,14 +30,18 @@ const BOOT_MAX_IDLE: Duration = Duration::from_secs(10);
 /// produced at all times.
 ///
 /// The returned Receiver can be awaited
-pub fn wait_for_partition_boot(config_store: Arc<RwConfigStore>, partition: u16) -> Receiver<()> {
+pub fn wait_for_partition_boot(
+    config_store: Arc<RwConfigStore>,
+    partition: u16,
+    shutdown: CancellationToken,
+) -> Receiver<bool> {
     let start = Instant::now();
-    let (boot_finished, boot_finished_rx) = oneshot::channel::<()>();
+    let (boot_finished, boot_finished_rx) = oneshot::channel::<bool>();
 
     tokio::spawn(async move {
         let mut interval = interval(Duration::from_millis(100));
 
-        loop {
+        while !shutdown.is_cancelled() {
             let tick = interval.tick().await;
             let elapsed = tick - start;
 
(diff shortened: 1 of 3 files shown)


src/config_waiter.rs
```diff
@@ -30,14 +30,18 @@ const BOOT_MAX_IDLE: Duration = Duration::from_secs(10);
 /// produced at all times.
 ///
 /// The returned Receiver can be awaited
-pub fn wait_for_partition_boot(config_store: Arc<RwConfigStore>, partition: u16) -> Receiver<()> {
+pub fn wait_for_partition_boot(
+    config_store: Arc<RwConfigStore>,
+    partition: u16,
+    shutdown: CancellationToken,
+) -> Receiver<bool> {
```
wedamija, line 37: Not sure if we should use something more explicit than a bool?

evanpurkhiser: Enum?

wedamija: Thinking about it and I'm not sure if it's much better? Just like returning a bool from a function really

evanpurkhiser: Yeah I'm indifferent.

I like enums because they're a bit more descriptive

like you could have

```
enum BootResult {
    Started,
    Canceled,
}
```

Then this returns a `Reciver<BootResult>` much more clear what it is

src/config_waiter.rs
```diff
@@ -64,19 +68,34 @@ pub fn wait_for_partition_boot(config_store: Arc<RwConfigStore>, partition: u16)
...
+            metrics::gauge!("config_consumer.partition_boot_time_ms", "partition" => partition.to_string())
+                .set(boot_time_ms as f64);
+            metrics::gauge!("config_consumer.partition_total_configs", "partition" => partition.to_string())
+                .set(total_configs as f64);
+        }
 
-        boot_finished.send(()).expect("Failed to report boot");
+        boot_finished
+            .send(!shutdown.is_cancelled())
+            .expect("Failed to report boot");
```
evanpurkhiser, line 98: So my thought here was to just early return, so right after the while loop you could do

```
if shutdown.is_cancelled() {
	boot_finished
			.send(false)
			.expect("Failed to report boot");
	return;
}
```

(or alternatively if we have the enum)

```rs
if shutdown.is_cancelled() {
	boot_finished
			.send(BootResult::Canceled)
			.expect("Failed to report boot");
	return;
}
```


Then we don't have to put any of the tracing stuff in an if block

src/config_waiter.rs
```diff
@@ -126,6 +146,40 @@ mod tests {
...
+
+        // Add a check config
+        config_store
+            .write()
+            .unwrap()
+            .add_config(Arc::new(CheckConfig::default()));
+
+        // Move time forward to the BOOT_MAX_IDLE. This will NOT mark the boot as complete sicne we
+        // just produced a config. We will need to wait another 400ms for it to complete
+        sleep(Duration::from_millis(200)).await;
```
evanpurkhiser, line 174: Probably don't need any of this stuff for this test, since you're just going to cancel it and we just want to validate that cancelation happens

evanpurkhiser (approved):

src/config_waiter.rs
wedamija: Yeah, I guess I wanted to simulator a partial startup and that nothing else goes wrong if we cancel

PR #158, opened 2024-08-30
chore: add GHA workflow to build the container

### LINKED ISSUE #157: Provide Docker image
This can be relatively easily implement with a Github Actions workflow.

TheDevMinerTV:
Fixes #157


aldy505:
cc @evanpurkhiser, please?

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+
+  schedule:
+    # We want the release to be at 9-10am Pacific Time
+    # We also want it to be 1 hour before the on-prem release
+    - cron: "0 17 15 * *"
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    name: "Release a new Relay version"
```
evanpurkhiser, line 21: This should be fixed

evanpurkhiser:
Going to forward this over to our open source release management team. Thanks!

hubertdeng123:
Looks like uptime-checker is in EA right now. Let's hold off on this for now until we can get this integrated into self-hosted

aldy505:
> Looks like uptime-checker is in EA right now. Let's hold off on this for now until we can get this integrated into self-hosted

@hubertdeng123 fun fact: I did get this integrated on my instance. https://github.com/getsentry/self-hosted/issues/3304

aldy505:
@TheDevMinerTV We would need to revisit this. Should refer to the symbolicator's way of building the Docker image and pushing to both GHCR and DockerHub. See this: https://github.com/getsentry/symbolicator/blob/master/.github/workflows/image.yml

It would be nice if we can build to arm64 too.

.craft.yml
```diff
@@ -0,0 +1,23 @@
...
+  name: github
+  config:
+    contexts:
+      - 'build-uptime-checker (sentryio)'
+
+targets:
+  - name: github
+  - id: release
+    name: docker
+    source: ghcr.io/getsentry/uptime-checker
```
aldy505, line 16: Different than other repos, I guess using GHCR as the default source is a better thing to do since it is guaranteed open source and guaranteed not to be tampered with.

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,74 @@
...
+          docker buildx build \
+              --platform linux/arm64,linux/amd64 \
+               $( [[ "${GITHUB_REF}" == "refs/heads/main" ]] && printf %s "--tag ${NIGHTLY_IMAGE}" ) \
+              --tag "${VERSIONED_IMAGE}" \
+              --file Dockerfile
+              --push
+              .
+
+  publish-to-dockerhub:
+    name: Publish uptime-checker to DockerHub
```
aldy505, line 44: After publishing to GHCR, you could remove the GCR build and just copy the GHCR image to GCR like what Relay did here https://github.com/getsentry/relay/blob/ac330795fdb9d00ce158f200c380b7b364100bd5/.github/workflows/ci.yml#L403-L408

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.80-alpine3.20 as builder
+FROM rust:1.80-alpine3.20 AS builder
```
aldy505, line 1: A nitpick, using lowercased "as" emit a warning on Docker.

aldy505 (commented):
cc @wedamija @evanpurkhiser

aldy505:
@wedamija @evanpurkhiser Any chance to take a look at this one?

JoshFerge (approved):

aldy505:
Any chance this will get merged before 15 March?

aldy505:
@wedamija @evanpurkhiser Anything?

.craft.yml
BYK: I think this should be done in 2 separate steps as currently we build on Google Cloud and the GitHub context you block on (defined above) is the Google Cloud Build status anyway. Let's go with that image first and then do the switch over across all repos.

```suggestion
    source: us-central1-docker.pkg.dev/sentryio/uptime-checker/image
```

.github/workflows/build-image.yaml
BYK: All this should be in a separate patch

.github/workflows/release.yml
```diff
@@ -0,0 +1,43 @@
...
+      version:
+        description: Version to release (optional)
+        required: false
+      force:
+        description: Force a release even when there are release-blockers (optional)
+        required: false
+
+  schedule:
+    # We want the release to be at 9-10am Pacific Time
+    # We also want it to be 1 hour before the on-prem release
```
BYK, line 15:
```suggestion
    # We also want it to be 1 hour before the self-hosted release
```

BYK (changes requested):
Rejecting to split the GHCR builds from the actual release set up.

.craft.yml
aldy505: @BYK How can we guarantee if the image on the GCR is being successfully built?

.github/workflows/build-image.yaml
aldy505: @BYK By "all this", did you mean removing the GCR build?

.craft.yml
BYK: @aldy505 that's what Craft does with the status check? GitHub status check would fail if that Google Cloud build fails which then would be picked up by Craft, cancelling the publishing?

.github/workflows/build-image.yaml
BYK: Yup, this entire job.

.craft.yml
aldy505: Ah! TIL.

aldy505:
@BYK Can you re-review today?

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,74 @@
...
+        run: docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+
+      - name: Build and push to GHCR
+        run: |
+          docker buildx build \
+              --platform linux/arm64,linux/amd64 \
+               $( [[ "${GITHUB_REF}" == "refs/heads/main" ]] && printf %s "--tag ${NIGHTLY_IMAGE}" ) \
+              --tag "${VERSIONED_IMAGE}" \
+              --file Dockerfile \
+              --push \
```
BYK, line 40: Less error prone: (see https://yaml-multiline.info/ for more info)
```suggestion
        run: >-
          docker buildx build
              --platform linux/arm64,linux/amd64
               $( [[ "${GITHUB_REF}" == "refs/heads/main" ]] && printf %s "--tag ${NIGHTLY_IMAGE}" )
              --tag "${VERSIONED_IMAGE}"
              --file Dockerfile
              --push
```

.github/workflows/build-image.yaml
aldy505: I don't see why not now? If we rely on the GCR build, we would need the `until ...` command that would take up a long time for the GHA to finish.

.github/workflows/build-image.yaml
```diff
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   build-image:
     name: Build and publish uptime-checker to GHCR
-    if: !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'
+    if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
```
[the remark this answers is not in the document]
aldy505: We could simply do
```yml
if: github.repository_owner == 'getsentry'
```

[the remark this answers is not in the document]
BYK: Not sure if that works with dependabot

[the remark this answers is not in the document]
BYK: Yeah it doesn't and we had to do something contrived over at Spotlight too: https://github.com/getsentry/spotlight/pull/702/files

.craft.yml
```diff
@@ -0,0 +1,23 @@
...
+
+targets:
+  - name: github
+  - id: release
+    name: docker
+    source: us-central1-docker.pkg.dev/sentryio/uptime-checker/image
+    target: getsentry/uptime-checker
+  - id: latest
+    name: docker
+    source: ghcr.io/getsentry/uptime-checker
```
BYK, line 20: I think these sources should match?

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+      - uses: docker/setup-qemu-action@v3
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Login to GHCR
+        run: docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+
+      - name: Build and push to GHCR
+        run: >-
+          docker buildx build
```
BYK, line 35: I really do not like us double building this image (once in Google Cloud and once in GHA). Looks like a big waste of resources. @asottile-sentry do you know if we have any plans to address this?

BYK (approved):
Not happy about the double building and need an answer for the different `source` parts in the Craft config file. Otherwise looks good, thanks a lot for getting through with this!

aldy505:
Any updates? It's been months now.

aldy505:
Anything? @evanpurkhiser @BYK @hubertdeng123

.github/workflows/build-image.yaml
mdtro, line 1: Let's pin the GitHub Actions to their respective SHAs instead of just the `@v1`, `@v3`, etc tags.

.github/workflows/release.yml
mdtro, line 1: Let's pin the GitHub Actions to their respective SHAs instead of just the `@v1`, `@v3`, etc tags.

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,74 @@
...
+  push:
+    branches:
+      - main
+      - release/**
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
```
mdtro, line 14: Set appropriate permissions for the jobs within the action. You can do this for all jobs, or within each specific job entry.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

evanpurkhiser:
Gonna go ahead and merge this and we can just iterate, sorry for blocking you forever on this @aldy505

PR #169, opened 2024-09-26
feat: Implemente RestrictedResolver for DNS resolution

evanpurkhiser:
This resolver allows a set of network ranges to be specified that will
error on resolution.


src/checker/http_checker.rs
```diff
@@ -327,6 +397,31 @@ mod tests {
...
+        assert_eq!(result.request_info.and_then(|i| i.http_status_code), None);
+
+        assert_eq!(
+            result.status_reason.as_ref().map(|r| r.status_type),
+            Some(CheckStatusReasonType::DnsError)
+        );
+        assert_eq!(
+            result.status_reason.map(|r| r.description),
+            Some("Failed to resolve address".to_string())
+        );
```
wedamija, line 422: Could be worth explicitly testing some bad ipv4 and ipv6 ips too

wedamija (approved):
As a follow up it'd be good to be able to make this an explicit failure reason

src/checker/http_checker.rs
```diff
@@ -17,6 +24,56 @@ use super::Checker;
...
+            Err(err) => {
+                let boxed_err: BoxError = Box::new(err);
+                Err(boxed_err)
+            }
+            Ok(mut addrs) => {
+                let is_restricted =
+                    addrs.any(|addr| ranges.iter().any(|ip_net| ip_net.contains(&addr.ip())));
+
+                if is_restricted {
+                    let boxed_err: BoxError = Box::from("Failed to resolve address".to_owned());
```
mdtro, line 66: Should we instead use an error indicating its a restricted IP? These might mix in with actual DNS resolution failures and make troubleshooting confusing in the future.

mdtro (approved):

Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "uptime-checker"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.79"
+rust-version = "1.80"
```
evanpurkhiser, line 5: Bumped to bring in LazyCell to construct the vector of restricted addresses

oioki (approved):

evanpurkhiser:
Superseded by #172

PR #174, opened 2024-09-30
ref: Sanitize generic failure error message url

evanpurkhiser:
If the URL included a token or something we wouldn't sanitize it, let's avoid including the URL

### DIFF
--- a/src/checker/http_checker.rs
+++ b/src/checker/http_checker.rs
@@ -165,7 +165,7 @@ impl Checker for HttpChecker {
                 } else {
                     CheckStatusReason {
                         status_type: CheckStatusReasonType::Failure,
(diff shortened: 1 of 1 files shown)


wedamija (approved):
Probably safer, although it might be good to improve our logging to attempt to fill in the url from the subscription, since i think this will remove it

evanpurkhiser:
Yeah, it doesn't look like we actually do any logging though 🙃

wedamija:
> Yeah, it doesn't look like we actually do any logging though 🙃

We log the result somewhere, I'm pretty sure I've used this url being part of the results to help figure things out in logs

evanpurkhiser:
Oh, this just removes it from the error though

PR #177, opened 2024-11-19
feat(uptime): bump sentry kafka schemas to latest

JoshFerge:
bump to latest so we get https://github.com/getsentry/sentry-kafka-schemas/pull/351

part of https://github.com/getsentry/team-uptime/issues/22


evanpurkhiser (approved):
afaik we don't actually use it, but 👍

JoshFerge:
> afaik we don't actually use it, but 👍

why not?

PR #189, opened 2024-12-19
feat(uptime): Update config schema and example tests

wedamija:
Bumping this to validate that the checker can handle the new config updates. I'll update the types in a later pr.


JoshFerge (approved):

evanpurkhiser (approved):
We'll update the config struct in the future right?

wedamija:
> We'll update the config struct in the future right?

Yeah, I'll make those changes when I get to updating the uptime-checker in general

PR #192, opened 2024-12-31
feat(uptime): Add region configuration to `CheckConfig`.

wedamija:
This adds `active_regions` and `region_schedule_mode` to `CheckConfig` and updates tests. These will be used to schedule checks when multiple regions are present.

### DIFF
--- a/src/types/check_config.rs
+++ b/src/types/check_config.rs
@@ -5,7 +5,7 @@ use serde_with::serde_as;
 use std::hash::{Hash, Hasher};
 use uuid::Uuid;
 
-use super::shared::RequestMethod;
+use super::shared::{RegionScheduleMode, RequestMethod};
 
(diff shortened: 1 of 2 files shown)


src/types/check_config.rs
```diff
@@ -59,6 +59,12 @@ pub struct CheckConfig {
     /// If we should allow sampling on the trace spans.
     #[serde(default)]
     pub trace_sampling: bool,
+
+    #[serde(default)]
+    pub active_regions: Option<Vec<String>>,
```
evanpurkhiser, line 64: We'll change these from Option in the future right?

evanpurkhiser (approved):

src/types/check_config.rs
wedamija: Ideally yeah, this should always be present

PR #193, opened 2024-12-31
feat(uptime): Add `region` to the Uptime service config.

wedamija:
This allows region to be specified at the checker level, so that the checker knows where it is running

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -14,6 +14,7 @@ use crate::config_store::{RwConfigStore, Tick};
 use crate::config_waiter::BootResult;
 use redis::{AsyncCommands, Client};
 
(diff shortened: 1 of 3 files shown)


src/scheduler.rs
```diff
@@ -43,6 +46,7 @@ async fn scheduler_loop(
     shutdown: CancellationToken,
     progress_key: String,
     redis_host: String,
+    _region: String,
```
evanpurkhiser, line 49: basically just not using it yet right

evanpurkhiser (approved):

src/scheduler.rs
wedamija: Yeah exactly, used in a followup

PR #194, opened 2024-12-31
feat(uptime): Make scheduler aware of regions, and able to skip checks when it's not a given region's turn.

wedamija:
Now that we're introducing multiple regions the scheduler needs to be aware of which region it's in, and of the regions in the configs. It can then only run checks for regions when it's its turn to do so.

To calculate whether it's time to run, we look at the current timestamp, and divide it by the interval size. This gets us the total number of time that interval has occurred. We then modulo that by the number of regions, to figure out which region's turn it is to run. Finally, with that number, we compare the current region to that and decide whether to run.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ tokio-stream = "0.1.15"
 redis = { version = "0.26.0", features = ["tokio-comp"] }
 ipnet = "2.10.0"
 hickory-resolver = "0.24.1"
+once_cell = "1.19.0"
 
 [patch.crates-io]
 rdkafka = { git = "https://github.com/fede1024/rust-rdkafka" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,6 @@ tokio-stream = "0.1.15"
 redis = { version = "0.26.0", features = ["tokio-comp"] }
 ipnet = "2.10.0"
 hickory-resolver = "0.24.1"
-once_cell = "1.19.0"
 
 [patch.crates-io]
 rdkafka = { git = "https://github.com/fede1024/rust-rdkafka" }
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -195,7 +195,7 @@ mod tests {
     #[traced_test]
     #[tokio::test(start_paused = true)]
     async fn test_scheduler() {
-                // We don't want these tests to run at the same time, since they access the external redis.
+        // We don't want these tests to run at the same time, since they access the external redis.
         let _guard = TEST_MUTEX.lock().await;
 
         let config = Config::default();
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -46,7 +46,7 @@ async fn scheduler_loop(
     shutdown: CancellationToken,
     progress_key: String,
     redis_host: String,
-    _region: String,
+    region: String,
 ) {
     let client = Client::open(redis_host.clone()).unwrap();
     let mut connection = client
@@ -93,7 +93,17 @@ async fn scheduler_loop(
         // TODO(epurkhiser): Check if we skipped any ticks If we did we should catch up on those.
 
         for config in configs {
-            results.push(queue_check(&executor_sender, tick, config));
+            if config.should_run(tick, region.clone()) {
+                results.push(queue_check(&executor_sender, tick, config));
+            } else {
+                tracing::debug!(%config.subscription_id, %tick, "scheduler.skipped_config");
+
+                metrics::counter!(
+                    "scheduler.skipped_region",
+                    "checker_region" => region.clone(),
+                )
+                .increment(1);
+            }
         }
 
         // Spawn a task to wait for checks to complete.
@@ -163,11 +173,13 @@ mod tests {
     use chrono::{Duration, Utc};
     use redis::{Client, Commands};
     use similar_asserts::assert_eq;
-    use std::sync::Arc;
+    use std::sync::{Arc, Mutex};
     use tokio::sync::{mpsc, oneshot};
     use tokio_util::sync::CancellationToken;
     use tracing_test::traced_test;
     use uuid::Uuid;
+    use once_cell::sync::Lazy;
+
 
     use super::run_scheduler;
 
@@ -180,10 +192,16 @@ mod tests {
             result::{CheckResult, CheckStatus},
         },
     };
+    use crate::types::shared::RegionScheduleMode;
+    static TEST_MUTEX: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
+
 
     #[traced_test]
     #[tokio::test(start_paused = true)]
     async fn test_scheduler() {
+        // We don't want these tests to run at the same time, since they access the external redis.
+        let _guard = TEST_MUTEX.lock().unwrap();
+
         let config = Config::default();
         let partition = 0;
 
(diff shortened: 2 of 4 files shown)


wedamija:
Note: This pr is a mess because it's based on multiple branches. I'll rebase once the others merge

Cargo.toml
```diff
@@ -37,6 +37,7 @@ tokio-stream = "0.1.15"
 redis = { version = "0.26.0", features = ["tokio-comp"] }
 ipnet = "2.10.0"
 hickory-resolver = "0.24.1"
+once_cell = "1.19.0"
```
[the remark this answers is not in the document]
evanpurkhiser:
```rs
use std::sync::{Mutex, OnceLock};

static TEST_MUTEX: OnceLock<Mutex<()>> = OnceLock::new();

fn main() {
    let mutex = TEST_MUTEX.get_or_init(|| Mutex::new(()));
    
    // Use the mutex directly
    let _lock = mutex.lock().unwrap();
    println!("Mutex is locked!");
}```

`OnceLock` I think?

File: src/types/check_config.rs [hunk 2]
```diff
@@ -91,17 +91,31 @@ impl CheckConfig {
...
             .collect()
     }
+
+    pub fn should_run(&self, tick: Tick, current_region: String) -> bool {
+        // Determines if this check should run in the current region at this current tick
+
+        // If region configs aren't present, assume we should always run this check
+        if self.active_regions.as_ref().is_none() || self.region_schedule_mode.is_none() {
+            return true;
+        }
```
evanpurkhiser, line 101: I think there's probably a more rusty way to write this to get the value into a variable otherwise return false, since now later you're having to do `unwrap`, but that unwrap isn't type checked. if you removed this guard here the compiler would still happily compile and just panic if it was None

evanpurkhiser:
```rs
if self.region_schedule_mode.is_none() {
  return true;
}

let active_regions = Some(self.active_regions.as_ref()) else {
  return true;
}

let running_region_idx = (tick.time().timestamp() / self.interval as i64)
    .checked_rem(active_regions.len() as i64)
    .unwrap();

self.active_regions.as_ref().unwrap()[running_region_idx as usize] == current_region
```

evanpurkhiser: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

Cargo.toml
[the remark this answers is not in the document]
untitaker: lazylock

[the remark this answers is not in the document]
evanpurkhiser: ah like this right

```rs
use std::sync::{Lazy, Mutex};

static TEST_MUTEX: Lazy<Mutex<()>> = Lazy::new(|| Mutex::new(()));
```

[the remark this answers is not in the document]
untitaker: I am not sure if you need LazyLock at all, since Mutex::new is const. try it out. Lazy might not work bc it's not Sync

src/scheduler.rs
```diff
@@ -93,7 +93,17 @@ async fn scheduler_loop(
         // TODO(epurkhiser): Check if we skipped any ticks If we did we should catch up on those.
```
evanpurkhiser, line 93: todo is probably outdated, unrealted

src/types/check_config.rs
```diff
@@ -91,17 +91,35 @@ impl CheckConfig {
             .map(|c| first_slot + (interval_secs * c))
             .collect()
     }
+
+    pub fn should_run(&self, tick: Tick, current_region: String) -> bool {
```
evanpurkhiser, line 95: can this borrow the region instead of needing to own it?

evanpurkhiser: Then you won't need to clone when calling this

src/scheduler.rs
```diff
@@ -173,17 +183,22 @@ mod tests {
...
             result::{CheckResult, CheckStatus},
         },
     };
+    static TEST_MUTEX: Mutex<()> = Mutex::new(());
 
     #[traced_test]
     #[tokio::test(start_paused = true)]
     async fn test_scheduler() {
+        // We don't want these tests to run at the same time, since they access the external redis.
+        let _guard = TEST_MUTEX.lock().unwrap();
```
evanpurkhiser, line 200: I wonder if there's actually a better way to do this. Let me post in the rust channel

src/types/check_config.rs
```diff
@@ -256,4 +274,232 @@ mod tests {
             all_slots.skip(15).step_by(60 * 5).collect::<Vec<usize>>()
         );
     }
+
+    #[test]
+    fn test_should_run() {
+        let tick = Tick::from_time(Utc::now());
```
evanpurkhiser, line 280: Is time frozen or something, how does this work 🤔

src/types/check_config.rs
```diff
@@ -256,4 +274,232 @@ mod tests {
...
+            "us_west".to_string()
+        ));
+        assert!(!multi_region_config_large_interval.should_run(
+            Tick::from_time(DateTime::from_timestamp(((60 * 60) * 13423) + (60 * 60), 0).unwrap()),
+            "us_east".to_string()
+        ));
+        assert!(!multi_region_config_large_interval.should_run(
+            Tick::from_time(DateTime::from_timestamp(((60 * 60) * 13423) + (60 * 60), 0).unwrap()),
+            "eu_west".to_string()
+        ));
```
evanpurkhiser, line 503: nit, I might turn this computation into a little helper function or something

```rs
fn test_time(offset: u32) -> DateTime {
  return DateTime::from_timestamp(((60 * 60) * 13423) + (60 * offset), 0).unwrap()
}
```

idk, might make it a bit easier to see whats diff between each call here

src/types/check_config.rs
```diff
@@ -91,17 +91,35 @@ impl CheckConfig {
...
-    use similar_asserts::assert_eq;
-    use uuid::{uuid, Uuid};
-
     use super::{CheckConfig, CheckInterval, RequestMethod};
+    use crate::config_store::Tick;
     use crate::types::check_config::MAX_CHECK_INTERVAL_SECS;
     use crate::types::shared::RegionScheduleMode;
+    use chrono::{DateTime, TimeDelta, Utc};
+    use similar_asserts::assert_eq;
+    use uuid::{uuid, Uuid};
```
evanpurkhiser, line 122: Ah your IDE is re-ordering these I think. All good, kinda unfortunate rn there's no way to keep these consistent

evanpurkhiser (approved):

src/types/check_config.rs
wedamija: Oh it just doesn't matter for the calls that use it, since they never check the tick

evanpurkhiser: Ahh I see

src/scheduler.rs
```diff
@@ -173,17 +181,22 @@ mod tests {
...
     use crate::manager::build_progress_key;
+    use crate::types::shared::RegionScheduleMode;
     use crate::{
         config_store::ConfigStore,
         types::{
             check_config::CheckConfig,
             result::{CheckResult, CheckStatus},
         },
     };
+    static TEST_MUTEX: Mutex<()> = Mutex::new(());
```
evanpurkhiser, line 192: lol this is easier huh

evanpurkhiser (approved):

src/types/check_config.rs
```diff
@@ -301,21 +301,21 @@ mod tests {
         };
         assert!(one_region_config.should_run(
             Tick::from_time(DateTime::from_timestamp(1, 0).unwrap()),
-            "us_west".to_string()
+            &"us_west".to_string()
```
evanpurkhiser, line 311: I think if you make the function take `&str` this would just be `"us_west"`

src/types/check_config.rs
```diff
@@ -328,99 +339,154 @@ mod tests {
...
-        ));
+        run_interval_test(
+            &multi_region_config,
+            TestInterval::OneMinute,
+            0,
+            TestInterval::OneSecond,
+            29,
+            vec!["us_west".to_string()],
+            vec!["eu_west".to_string(), "us_east".to_string()],
+        );
```
evanpurkhiser, line 384: pretty small nit, but I think I do kind of wish it took a struct so we could see what the names of each of these parameters are. but I'm sure some IDEs show ghost labels for the parameters.

indifferent

PR #196, opened 2025-01-06
feat(uptime): Add region to uptime results and stats

wedamija:
We need to start returning regions in our results. Also including them in our various stats so that we can break things down by uptime region if necessary

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -70,8 +70,9 @@ pub fn run_executor(
     tracing::info!("executor.starting");
 
     let (sender, reciever) = mpsc::unbounded_channel();
-    let executor =
-        tokio::spawn(async move { executor_loop(concurrency, checker, producer, reciever, region).await });
+    let executor = tokio::spawn(async move {
+        executor_loop(concurrency, checker, producer, reciever, region).await
+    });
 
     (sender, executor)
 }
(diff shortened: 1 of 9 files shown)


src/check_executor.rs
```diff
@@ -100,13 +103,15 @@ async fn executor_loop(
     checker: Arc<impl Checker + 'static>,
     producer: Arc<impl ResultsProducer + 'static>,
     reciever: UnboundedReceiver<ScheduledCheck>,
+    region: String,
```
evanpurkhiser, line 106: Is it actually possible to make this a `&str` then we wouldn't need to clone it around quite as much I think

src/types/result.rs
```diff
@@ -96,6 +96,8 @@ pub struct CheckResult {
 
     /// Information about the check request made. Will be empty if the check was missed
     pub request_info: Option<RequestInfo>,
+
+    pub region: String,
```
evanpurkhiser, line 100: Might as well keep up the comments
```suggestion
    /// Region slug that produced the check result
    pub region: String,
```

src/check_executor.rs
```diff
@@ -194,15 +200,17 @@ fn record_result_metrics(result: &CheckResult) {
...
     metrics::counter!(
-        "check_result.processed",
-        "status" => status_label,
-        "failure_reason" => failure_reason.unwrap_or("ok"),
-        "status_code" => status_code,
+           "check_result.processed",
+           "status" => status_label,
+           "failure_reason" => failure_reason.unwrap_or("ok"),
+           "status_code" => status_code,
+           "checker_region" => result.region.clone(),
```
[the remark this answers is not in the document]
wedamija: Hmm, I think my formatter added some tabs...

evanpurkhiser (approved):

PR #197, opened 2025-01-11
feat(uptime): Make sure result and trace ids are consistent for a scheduled check

wedamija:
We want to ensure that we produce the same ids for the same check. Each scheduled check is uniquely identified via the subscription id and scheduled time. Having these ids be consistent makes it easy for us to de-dupe data if we have duplicates due to a deploy or other issue.

### DIFF
--- a/src/checker/http_checker.rs
+++ b/src/checker/http_checker.rs
@@ -1,7 +1,7 @@
 use chrono::{TimeDelta, Utc};
 use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
 use reqwest::{Client, ClientBuilder, Response};
-use sentry::protocol::{SpanId, TraceId};
+use sentry::protocol::SpanId;
 use std::error::Error;
 use tokio::time::Instant;
 use uuid::Uuid;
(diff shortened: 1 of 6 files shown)


src/types/result.rs
```diff
@@ -74,7 +74,8 @@ pub struct CheckResult {
     pub status_reason: Option<CheckStatusReason>,
 
     /// Trace ID associated with the check-in made
-    pub trace_id: TraceId,
+    #[serde(serialize_with = "uuid_simple")]
+    pub trace_id: Uuid,
```
wedamija, line 78: I couldn't instantiate `TraceId` correctly since it's private, but I don't think it adds too much here since it's basically just a uuid

src/checker/http_checker.rs
```diff
@@ -15,6 +15,8 @@ use crate::types::{
 use super::ip_filter::is_external_ip;
 use super::Checker;
 
+pub const CHECKER_RESULT_NAMESPACE: Uuid = Uuid::from_u128(0x67f0b2d5_e476_4f00_9b99_9e6b95c3b7e3);
```
wedamija, line 18: Just a random uuid as the base for our v5 uuid namespace

JoshFerge (approved):
lgtm

PR #198, opened 2025-01-16
feat: better error formatting

JoshFerge:
- [x] Format timed out errors
- [x] Format ssl errors


Dockerfile.localdev
```diff
@@ -0,0 +1,30 @@
+FROM rust:1.80-alpine3.20 as builder
```
JoshFerge, line 1: since errors are different by platform, made this local docker file so we can quickly develop on linux. i'll add a comment at the top with instructions on how to use

src/checker/http_checker.rs
```diff
@@ -489,6 +533,238 @@ mod tests {
         let trace_header_sampling = make_trace_header(&config_with_sampling, &trace_id, span_id);
         assert_eq!(trace_header_sampling, format!("{}-{}", trace_id, span_id));
     }
-    // TODO: Figure out how to simulate a DNS failure
-    // assert_eq!(check_domain(&client, "https://hjkhjkljkh.io/".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
+
+    #[tokio::test]
+    // only test on mac
+    #[cfg(target_os = "macos")]
```
JoshFerge, line 539: i think i'm just gonna delete these, as they don't really help anyone

src/checker/http_checker.rs
```diff
@@ -489,6 +533,238 @@ mod tests {
...
+                    .map(|r| r.description.clone())
+                    .unwrap()
+            );
+        }
+    }
+    
+
+    #[tokio::test]
+    // #[cfg(target_os = "linux")]
+    async fn test_ssl_errors_linux() {
```
JoshFerge, line 656: it is probably not prudent to merge this test in as is. https://github.com/chromium/badssl.com

i feel like it would take a while to figure out how to mock this correctly or self-host the bad-ssl website for tests, so maybe its best to just comment out?

src/types/result.rs
```diff
@@ -30,6 +30,7 @@ pub enum CheckStatusReasonType {
     Timeout,
     DnsError,
     Failure,
+    SslError,
```
evanpurkhiser, line 33: If we add to this we'll need to update the schemas

https://github.com/getsentry/sentry-kafka-schemas/blob/main/schemas/uptime-results.v1.schema.json#L6-L11

evanpurkhiser: Even just having the better error message now is going to be helpful!

src/checker/http_checker.rs
```diff
@@ -489,6 +526,112 @@ mod tests {
...
         assert_eq!(trace_header_sampling, format!("{}-{}", trace_id, span_id));
     }
-    // TODO: Figure out how to simulate a DNS failure
-    // assert_eq!(check_domain(&client, "https://hjkhjkljkh.io/".to_string()).await, CheckResult::FAILURE(FailureReason::DnsError("failed to lookup address information: nodename nor servname provided, or not known".to_string())));
+
+    #[tokio::test]
+    #[cfg(target_os = "linux")]
+    async fn test_ssl_errors_linux() {
+        // Helper function to create various bad certificates
+        fn create_bad_cert(cert_type: &str) -> (Vec<u8>, Vec<u8>) {
```
evanpurkhiser, line 534: maybe make a little enum because why not

JoshFerge: in the tests? is that really necessary?

evanpurkhiser: I mean I think test code is just as important as production code. There's no overhead to adding an enum or anything like that. I just tend to really dislike magic strings.

```rs
enum TestCertType {
    Expired,
    WrongHost,
	SelfSigned,
}
```

anyway more of a nit, not too important

evanpurkhiser (approved):

src/checker/http_checker.rs
JoshFerge: i've updated the tests to use an Enum.

Dockerfile.localdev
```diff
@@ -0,0 +1,43 @@
...
+# It is not used for production. We need this because certain error messages
+# are Linux specific so we have to run the binary on Linux to get the correct
+# error messages.
+
+# HOW TO USE:
+# 1. Build the binary: `docker build -f Dockerfile.localdev -t my-uptime-checker`
+# 2. Run the container with a mounted volume for src: `docker run -it -v $(pwd)/src:/app/src --entrypoint /bin/sh my-uptime-checker`
+# e. in the container you can do `cargo test`, etc. and updating the src directory on your host will reflect in the container,
+# so you can test your changes without having to rebuild the container.
+##
```
wedamija, line 12: Thanks for this, helpful to both have this container and easy instructions on how to use it.

wedamija (approved):

PR #199, opened 2025-01-21
feat(uptime): Refactor config provider code

wedamija:
This refactors the config provider code so that we can introduce a new redis config provider. I mostly moved code around, it's probably not worthwhile fully building an abstraction here, since we'll remove Kafka soon.


src/app/config.rs
```diff
@@ -56,6 +62,8 @@ pub struct Config {
     /// The topic to load [`CheckConfig`]s from.
     pub configs_kafka_topic: String,
 
+    pub config_provider_mode: ConfigProviderMode,
```
evanpurkhiser, line 65: Let's add a comment for consistency

evanpurkhiser (approved):

PR #200, opened 2025-01-22
feat: allow internal ips for self-hosted instances

aldy505:
For self-hosted use cases, we will be checking against private IP address space, most of the times. Since most of the self-hosted deployment are done in a private (and mostly air-gapped) environment.

This PR introduces `allow_internal_ips` on the configuration that hopefully does not break with existing running instances.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -61,6 +61,9 @@ pub struct Config {
 
     /// The region that this checker is running in
     pub region: String,
+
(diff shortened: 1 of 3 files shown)


wedamija (approved):
Thanks for the contribution, this looks good to me. I'll merge shortly

evanpurkhiser (approved):

PR #202, opened 2025-01-22
fix: add connection refused handling

JoshFerge:
- [x] adds an additional clause for `Connection refused` in `connection_error` parsing.
- [x] adds a new function `hyper_error` to correctly format hyper errors that are returned from reqwest. a test is added to test this behavior with an `IncompleteMessage`. it is a different error on linux than mac, so put it behind that flag.


JoshFerge:
need to adjust this as this is different on linux.

wedamija (approved):

PR #205, opened 2025-01-22
chore(uptime): Make redis use in tests safer.

wedamija:
This generalizes our Redis test lock so that we can easily apply it to other tests. Also flushes the redis db when we're done to avoid test pollution.


Cargo.toml
```diff
@@ -43,7 +43,10 @@ tokio-rustls = "0.26.1"
 rustls = "0.23.21"
 rcgen = "0.11"
 time = "0.3"
-
+ctor = "0.2"
+syn = { version = "2.0", features = ["full"] }
+quote = "1.0"
+once_cell = "1.19.0"
```
evanpurkhiser, line 49: You shouldn't need once cell, this is built into rust now.

evanpurkhiser: Oh looks like you're not even using it?

Cargo.toml
```diff
@@ -43,7 +43,10 @@ tokio-rustls = "0.26.1"
 rustls = "0.23.21"
 rcgen = "0.11"
 time = "0.3"
-
+ctor = "0.2"
+syn = { version = "2.0", features = ["full"] }
+quote = "1.0"
```
evanpurkhiser, line 48: Does this dependency need to be here since it's in the `redis-test-macro` crate?

src/main.rs
```diff
@@ -30,3 +32,18 @@ pub fn main() {
...
+    use crate::app::config::Config;
+
+    #[ctor::dtor]
+    fn cleanup() {
+        let config = Config::default();
+        let client = Client::open(config.redis_host).unwrap();
+        let mut conn = client.get_connection().unwrap();
+        let _: () = redis::cmd("FLUSHDB").query(&mut conn).unwrap();
+    }
+}
```
evanpurkhiser, line 49: nit of nits: Missing trailing newline

src/main.rs
```diff
@@ -1,3 +1,4 @@
+#![deny(clippy::disallowed_types)]
```
evanpurkhiser, line 1: What this for?

src/main.rs
```diff
@@ -30,3 +32,18 @@ pub fn main() {
...
+mod test {
+    use redis::{Client};
+    use crate::app::config::Config;
+
+    #[ctor::dtor]
+    fn cleanup() {
+        let config = Config::default();
+        let client = Client::open(config.redis_host).unwrap();
+        let mut conn = client.get_connection().unwrap();
+        let _: () = redis::cmd("FLUSHDB").query(&mut conn).unwrap();
```
evanpurkhiser, line 47: Wouldn't mind a comment here for what this is doing

evanpurkhiser (approved):

PR #207, opened 2025-01-23
feat(metrics): Record queue size of executor as a metric

evanpurkhiser:
This will allow us to monitor how many checks are being executed concurrently over time. Since we have a concurrency limit this will also help us understand when we're getting close to that limit

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -1,3 +1,4 @@
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
 use chrono::{TimeDelta, Utc};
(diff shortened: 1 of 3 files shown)


JoshFerge (approved):
seems sensible to me

src/check_executor.rs
```diff
@@ -91,15 +108,18 @@ pub fn queue_check(
         config,
         resolve_tx,
     };
-    sender
+    check_sender
+        .sender
         .send(scheduled_check)
         .expect("Failed to queue ScheduledCheck");
+    check_sender.queue_size.fetch_add(1, Ordering::SeqCst);
```
evanpurkhiser, line 115: Using sequential consistency for all the memory ordering operations here based on the answers in this question https://stackoverflow.com/questions/9553591/c-stdatomic-what-is-stdmemory-order-and-how-to-use-them

wedamija (approved):

PR #208, opened 2025-01-23
fix(build): pull docker image before building

JoshFerge:
looking at our docker builds and why they're so slow, i noticed its not using the docker image cache, so its just building from scratch every time.  [link](https://console.cloud.google.com/cloud-build/builds/cf7b799e-e730-4d9b-8dce-1ec966367953?project=345757944225&invt=AbnoWw&inv=1)

according to these docs you have to pull the image for cacheing to work. 
https://cloud.google.com/build/docs/optimize-builds/speeding-up-builds

will see if the build on this branch is faster and if so, can merge.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM rust:1.80-alpine3.20 as builder
 
-ARG UPTIME_CHECKER_GIT_REVISION
-ENV UPTIME_CHECKER_GIT_REVISION=$UPTIME_CHECKER_GIT_REVISION
+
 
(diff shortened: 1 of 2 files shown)


JoshFerge:
actually no. i think the reason is `UPTIME_CHECKER_GIT_REVISION` is changed every build, and this is at the top of the `Dockerfile`, so the whole image has to rebuild.

evanpurkhiser (approved):

JoshFerge:
still not working, need to tweak the rust build stuff some more

PR #210, opened 2025-01-23
feat(uptime): Add `RedisConfigProvider`

wedamija:
This adds a `RedisConfigProvider` to manage configs via Redis. We still need to hook this up to the manager, have a way to receive partition assignments, and start monitoring for updates.

### DIFF
--- a/src/types/check_config.rs
+++ b/src/types/check_config.rs
@@ -1,8 +1,8 @@
 use super::shared::{RegionScheduleMode, RequestMethod};
 use crate::config_store::Tick;
 use chrono::TimeDelta;
-use serde::Deserialize;
-use serde_repr::Deserialize_repr;
+use serde::{Serialize, Deserialize};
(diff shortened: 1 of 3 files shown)


src/types/check_config.rs
```diff
@@ -26,7 +26,7 @@ pub const MAX_CHECK_INTERVAL_SECS: usize = CheckInterval::SixtyMinutes as usize;
 
 /// The CheckConfig represents a configuration for a single check.
 #[serde_as]
-#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
evanpurkhiser, line 29: need this for tests?

evanpurkhiser (approved):

src/types/check_config.rs
wedamija: Yeah, makes it easier to serialize configs and do a full test

PR #212, opened 2025-01-24
feat(uptime): Add redis config updater

wedamija:
This adds a method that will regularly poll redis for updates to the configs.

Note that I  use a dict here for the updates - since we're killing services and then bringing them back up, there's no need to co-ordinate here. Using a dict allows us to de-dupe updates so that only the last one will be processed for a given config.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -79,6 +79,9 @@ pub struct Config {
     /// Which config provider to use to load configs into memory
     pub config_provider_mode: ConfigProviderMode,
 
+    // How frequently to poll redis for config updates when using the redis config provider
+    pub config_provider_redis_update_ms: u64,
+
     /// The general purpose redis node to use with this service
     pub redis_host: String,
 
@@ -107,6 +110,7 @@ impl Default for Config {
             configs_kafka_cluster: vec![],
             configs_kafka_topic: "uptime-configs".to_owned(),
             config_provider_mode: ConfigProviderMode::Kafka,
+            config_provider_redis_update_ms: 1000,
             redis_host: "redis://127.0.0.1:6379".to_owned(),
             region: "default".to_owned(),
             allow_internal_ips: false,
@@ -195,6 +199,7 @@ mod tests {
                     results_kafka_topic: "uptime-results".to_owned(),
                     configs_kafka_topic: "uptime-configs".to_owned(),
                     config_provider_mode: ConfigProviderMode::Kafka,
+                    config_provider_redis_update_ms: 1000,
                     redis_host: "redis://127.0.0.1:6379".to_owned(),
                     region: "default".to_owned(),
                     allow_internal_ips: false,
@@ -226,6 +231,7 @@ mod tests {
                 "10.0.0.1,10.0.0.2:7000",
             );
             jail.set_env("UPTIME_CHECKER_CONFIG_PROVIDER_MODE", "kafka");
+            jail.set_env("UPTIME_CHECKER_CONFIG_PROVIDER_REDIS_UPDATE_MS", "2000");
             jail.set_env("UPTIME_CHECKER_STATSD_ADDR", "10.0.0.1:1234");
             jail.set_env("UPTIME_CHECKER_REDIS_HOST", "10.0.0.3:6379");
             jail.set_env("UPTIME_CHECKER_REGION", "us-west");
@@ -257,6 +263,7 @@ mod tests {
                     results_kafka_topic: "uptime-results".to_owned(),
                     configs_kafka_topic: "uptime-configs".to_owned(),
                     config_provider_mode: ConfigProviderMode::Kafka,
+                    config_provider_redis_update_ms: 2000,
                     redis_host: "10.0.0.3:6379".to_owned(),
                     region: "us-west".to_owned(),
                     allow_internal_ips: true,
--- a/src/check_config_provider/redis_config_provider.rs
+++ b/src/check_config_provider/redis_config_provider.rs
@@ -6,6 +6,13 @@ use tokio_util::sync::CancellationToken;
 
 use crate::{app::config::Config, manager::Manager, types::check_config::CheckConfig};
 
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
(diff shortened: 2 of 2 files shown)


src/app/config.rs
```diff
@@ -79,6 +79,9 @@ pub struct Config {
     /// Which config provider to use to load configs into memory
     pub config_provider_mode: ConfigProviderMode,
 
+    // How frequently to poll redis for config updates when using the redis config provider
```
evanpurkhiser, line 82:
```suggestion
    /// How frequently to poll redis for config updates when using the redis config provider
```

https://doc.rust-lang.org/reference/comments.html#doc-comments

src/check_config_provider/redis_config_provider.rs
```diff
@@ -23,25 +30,69 @@ impl RedisPartition {
     }
 }
 
+/// The action to perform when receiving a ConfigUpdate
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ConfigUpdateAction {
+    Add,
+    Delete,
+}
```
evanpurkhiser, line 38: Oh I was thinking the updates would basically be a "upsert" or "delete".

src/check_config_provider/redis_config_provider.rs
```diff
@@ -23,25 +30,69 @@ impl RedisPartition {
...
+
+/// The CheckConfig represents a configuration for a single check.
```
evanpurkhiser, line 40: Wrong comment

src/check_config_provider/redis_config_provider.rs
wedamija: Yeah, upsert would be a better term here, I'll change that

src/check_config_provider/redis_config_provider.rs
```diff
@@ -88,6 +138,94 @@ impl RedisConfigProvider {
...
+                let mut pipe = redis::pipe();
+                // We fetch all updates from the list and then delete the key. We do this
+                // atomically so that there isn't any chance of a race
+                let (config_adds, config_deletes): (Vec<_>, Vec<_>) = pipe
+                    .atomic()
+                    .hvals(&partition.update_key)
+                    .del(&partition.update_key)
+                    .query_async::<(Vec<Vec<u8>>, ())>(&mut conn)
+                    .await
+                    .unwrap()
```
evanpurkhiser, line 168: Is the unwrap safe here?

src/check_config_provider/redis_config_provider.rs
```diff
@@ -88,6 +138,94 @@ impl RedisConfigProvider {
...
+        let mut interval = interval(self.check_interval);
+
+        while !shutdown.is_cancelled() {
+            let _ = interval.tick().await;
+
+            for partition in partitions.iter() {
+                let mut pipe = redis::pipe();
+                // We fetch all updates from the list and then delete the key. We do this
+                // atomically so that there isn't any chance of a race
+                let (config_adds, config_deletes): (Vec<_>, Vec<_>) = pipe
```
evanpurkhiser, line 162: damn that's a clean chain

evanpurkhiser: Let's make sure to rename these variables too `config_upserts`

src/check_config_provider/redis_config_provider.rs
wedamija: Probably not... I guess if we error, we should log an error and continue

src/check_config_provider/redis_config_provider.rs
```diff
@@ -88,6 +138,94 @@ impl RedisConfigProvider {
...
+                    .0 // Get just the LRANGE results
+                    .iter()
+                    .map(|payload| {
+                        rmp_serde::from_slice::<ConfigUpdate>(payload).map_err(|err| {
+                            tracing::error!(?err, "config_consumer.invalid_config_message");
+                            err
+                        })
+                    })
+                    .filter_map(Result::ok)
+                    .partition(|cu| cu.action == ConfigUpdateAction::Add);
```
evanpurkhiser, line 178: I sort of dislike that if we added another action it would bucket them into the second partition.

evanpurkhiser:
```suggestion
                    .partition(|cu| cu.action {
								        ConfigUpdateAction::Add => true,
								        ConfigUpdateAction::Delete => false,
								        // If we add a new variant, the compiler will now force us to handle it here
								    });
```

src/check_config_provider/redis_config_provider.rs
```diff
@@ -88,6 +138,94 @@ impl RedisConfigProvider {
...
+                        .get_service(partition.partition)
+                        .get_config_store()
+                        .write()
+                        .unwrap()
+                        .remove_config(config_delete.subscription_id);
+                    tracing::debug!(
+                        %config_delete.subscription_id,
+                        "config_consumer.config_removed"
+                    );
+                });
```
evanpurkhiser, line 191: I think you could just use a regular `for` loop here instead of the for each actually

wedamija: Yeah... It's not too clear to me which is prefferable

wedamija:
Just realised I need to add a bunch of telemetry, I'll add that in after lunch

src/check_config_provider/redis_config_provider.rs
```diff
@@ -88,6 +138,103 @@ impl RedisConfigProvider {
...
+                        })
+                    })
+                    .filter_map(Result::ok)
+                    .fold((vec![], vec![]), |(mut upserts, mut deletes), update| {
+                        match update.action {
+                            ConfigUpdateAction::Upsert => upserts.push(update),
+                            ConfigUpdateAction::Delete => deletes.push(update),
+                        }
+                        (upserts, deletes)
+                    });
```
evanpurkhiser, line 187: this is fine too. I actually didn't mind the partition as long as we had the match

evanpurkhiser (approved):

PR #213, opened 2025-01-24
feat(uptime): Add config values for pod name, total pod count and redis partitions

wedamija:
We'll use these to determine which partitions a checker is assigned.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -96,10 +96,10 @@ pub struct Config {
     pub allow_internal_ips: bool,
 
     /// Name of the pod that is running
-    pub pod_name: String,
+    pub checker_id: String,
 
-    /// Total number of uptime checker pods running
-    pub total_pods: usize,
+    /// Total number of uptime checkers running
+    pub total_checkers: usize,
 }
 
 impl Default for Config {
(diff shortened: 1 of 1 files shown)


src/app/config.rs
```diff
@@ -90,6 +94,12 @@ pub struct Config {
 
     /// Allow uptime checks against internal IP addresses
     pub allow_internal_ips: bool,
+
+    /// Name of the pod that is running
+    pub pod_name: String,
```
evanpurkhiser, line 99: wait do you need this, can't you just get `hostname`?

evanpurkhiser: What's this actually for

wedamija: Discussed on slack, will change to `checker_id`

src/app/config.rs
```diff
@@ -90,6 +94,12 @@ pub struct Config {
 
     /// Allow uptime checks against internal IP addresses
     pub allow_internal_ips: bool,
+
+    /// Name of the pod that is running
+    pub checker_id: String,
```
evanpurkhiser, line 99: Left this comment somewherer else, but wondiering if we should do the id deserialization here.

Could deserialize into a struct like `CheckerId` that includes the numeric value

evanpurkhiser (approved):

src/app/config.rs
wedamija: I'll do this in https://github.com/getsentry/uptime-checker/pull/214

PR #214, opened 2025-01-25
feat(uptime): Add function to determine which partitions a checker owns

wedamija:
This adds `determine_owned_partitions`, which produces a set of partition numbers that this checker owns based on total checkers, total partitions and this checker's checker number.

Simplifies the interface to calling run_config_provider since we can determine partitions via config rather than passing in.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -248,7 +248,10 @@ mod tests {
             );
             jail.set_env("UPTIME_CHECKER_CONFIG_PROVIDER_MODE", "kafka");
             jail.set_env("UPTIME_CHECKER_CONFIG_PROVIDER_REDIS_UPDATE_MS", "2000");
-            jail.set_env("UPTIME_CHECKER_CONFIG_PROVIDER_REDIS_TOTAL_PARTITIONS", "32");
+            jail.set_env(
+                "UPTIME_CHECKER_CONFIG_PROVIDER_REDIS_TOTAL_PARTITIONS",
+                "32",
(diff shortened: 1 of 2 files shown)


src/check_config_provider/redis_config_provider.rs
```diff
@@ -339,6 +338,28 @@ pub fn run_config_provider(
...
+pub fn determine_owned_partitions(config: &Config) -> HashSet<u16> {
+    // Determines which partitions this checker owns based on number of partitions,
+    // number of checkers and checker number
+    let checker_number: u16 = config
+        .checker_id
+        .split('-')
+        .last()
+        .expect("checker_id should be in format <name>-<checker_number>")
+        .parse()
+        .expect("checker_id does not contain a valid checker_number");
```
evanpurkhiser, line 350:
```suggestion
        .expect("checker_id should contain a valid checker_number");
```

keeps with the "expect <sentence that follows expect>"

src/app/config.rs
```diff
@@ -99,7 +99,7 @@ pub struct Config {
     pub checker_id: String,
```
evanpurkhiser, line 99: I wonder if it would actually make sense to move the number parsing into the deserialization (via a custom deserializer)

Since then you could produce an error message here in the config bit. Do we need the "full name" of the checker.

alternatively, why not jsut have the checker_id be just a number?

evanpurkhiser (approved):

PR #216, opened 2025-01-27
feat: add vector result producer

JoshFerge:
- [x] adds a new producer type, `VectorProducer`. This producer opens a channel and creates a background worker which will send check results to vector's http interface in batches, and does a final flush of all results when shut down. (TODO: make batch size configurable).
- [x] allow for configuration of producer
- [x] manage vector producer join handler in manage. this part is quite messy right now, and could use some help cleaning it up.

TODO: handle failure modes, determine how to handle if/when vector goes down (should we cap the size of the channel?)

### DIFF
--- a/src/manager.rs
+++ b/src/manager.rs
@@ -6,7 +6,10 @@ use std::{
     collections::{HashMap, HashSet},
     sync::{Arc, RwLock},
 };
-use tokio::sync::{mpsc::{self, UnboundedSender}, oneshot};
+use tokio::sync::{
+    mpsc::{self, UnboundedSender},
+    oneshot,
+};
 use tokio::task::JoinHandle;
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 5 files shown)


src/producer.rs
```diff
@@ -15,8 +16,13 @@ pub enum ExtractCodeError {
...
     Schema(#[from] SchemaError),
+    #[error(transparent)]
+    VectorRequestError(#[from] reqwest::Error),
+    #[error("Vector request failed with status: {0}")]
+    VectorRequestStatusError(reqwest::StatusCode),
 }
 
 pub trait ResultsProducer: Send + Sync {
     fn produce_checker_result(&self, result: &CheckResult) -> Result<(), ExtractCodeError>;
+
```
evanpurkhiser, line 27:
```suggestion
```

not sure why rustfmt didn't get this, weird

src/manager.rs
```diff
@@ -108,26 +109,40 @@ impl Manager {
...
-        // XXX: Executor will shutdown once the sender goes out of scope. This will happen once all
-        // referneces of the Sender (executor_sender) are dropped.
-        let (executor_sender, executor_join_handle) = run_executor(
-            config.checker_concurrency,
-            checker,
-            producer,
-            config.region.clone(),
-        );
+        let (executor_sender, (executor_join_handle, results_worker, results_shutdown_sender)) = match &config.producer_mode {
+            ProducerMode::Vector => {
```
JoshFerge, line 113: basically the problem here is that i can't simply set the producer equal to something, as the trait isn't sized? so have to do this weird match right now where we have to duplicate logic. i'm not quite sure what the fix is for that.

src/manager.rs
```diff
@@ -165,6 +180,10 @@ impl Manager {
             Box::pin(async move {
                 shutdown_signal.cancel();
                 consumer_join_handle.await.expect("Failed to stop consumer");
+                
+                results_worker.await.expect("Failed to stop vector worker");
+                results_shutdown_sender.send(()).expect("Failed to send shutdown signal");
```
JoshFerge, line 185: not sure if this line is necessary, i'm not sure if we have to send a shutdown signal here as we don't for our other join handles?

src/producer/vector_producer.rs
```diff
@@ -0,0 +1,405 @@
+use crate::producer::{ExtractCodeError, ResultsProducer};
+use crate::types::result::CheckResult;
+use reqwest::Client;
+use sentry_kafka_schemas::Schema;
+use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
+
+const VECTOR_BATCH_SIZE: usize = 2;
```
JoshFerge, line 9: TODO add this to config

src/manager.rs
evanpurkhiser: We send the signal right above here, line 181

src/producer/vector_producer.rs
```diff
@@ -0,0 +1,396 @@
...
+                    .post(&endpoint)
+                    .header("Content-Type", "application/json")
+                    .body(body)
+                    .send()
+                    .await
+                {
+                    tracing::error!(error = ?e, "Failed to send final batch request to Vector");
+                } else {
+                    tracing::debug!("Successfully sent final batch request");
+                }
```
evanpurkhiser, line 101: can you pull this logic out into a reusable function actually?

src/producer/vector_producer.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+
+            tracing::info!("Vector producer worker shutting down");
+        })
+    }
+}
+
+impl Drop for VectorResultsProducer {
+    fn drop(&mut self) {
+        // No longer need to handle shutdown in Drop since the worker is managed externally
```
evanpurkhiser, line 111: Is this still neded?

src/producer/vector_producer.rs
```diff
@@ -0,0 +1,396 @@
...
+    fn produce_checker_result(&self, result: &CheckResult) -> Result<(), ExtractCodeError> {
+        let json = serde_json::to_vec(result)?;
+        self.schema.validate_json(&json)?;
+
+        // Send the serialized result to the worker task
+        if self.sender.send(json).is_err() {
+            tracing::error!("Failed to send result to Vector worker - channel closed");
+            return Err(ExtractCodeError::VectorRequestStatusError(
+                reqwest::StatusCode::INTERNAL_SERVER_ERROR,
+            ));
```
evanpurkhiser, line 125: This feels strange. send might fail because the channel is closed why are we producing a status code internal server error? I would expect a different error type.

src/producer.rs
```diff
@@ -15,6 +16,10 @@ pub enum ExtractCodeError {
     Producer(#[from] ProducerError),
     #[error(transparent)]
     Schema(#[from] SchemaError),
+    #[error(transparent)]
+    VectorRequestError(#[from] reqwest::Error),
+    #[error("Vector request failed with status: {0}")]
+    VectorRequestStatusError(reqwest::StatusCode),
```
evanpurkhiser, line 22: I think we could name this differently, since how this is used right now is jsut for when the channel is closed

src/producer/vector_producer.rs
```diff
@@ -0,0 +1,396 @@
...
+
+pub struct VectorResultsProducer {
+    schema: Schema,
+    sender: UnboundedSender<Vec<u8>>,
+}
+
+impl VectorResultsProducer {
+    pub fn new(topic_name: &str) -> (Self, JoinHandle<()>) {
+        Self::new_with_endpoint(topic_name, "http://localhost:8020".to_string())
+    }
```
evanpurkhiser, line 18: IMO this should not be part of the public interface since it's just for testing. Maybe rename `new_with_endpoint` to just `new` and have it always take the vector endpoint.

Alternatively you can setup a public `VectorResultsProducerConfig` struct hat has these two things in it

evanpurkhiser (approved):

PR #221, opened 2025-01-29
feat(uptime): add new failure types

JoshFerge:
- [x] add connection_error and `tls_error` failure types. need https://github.com/getsentry/sentry-kafka-schemas/pull/377 rolled out to sentry and snuba before merging
- [x] log fallthrough errors, as they should not occur, and if we do, we should implement code to catch them.

I also considered making `connection_error` split into `connection_reset` and `connection_refused`. thoughts?

### DIFF
--- a/Dockerfile.localdev
+++ b/Dockerfile.localdev
@@ -30,6 +30,7 @@ WORKDIR /app
 RUN mkdir -p /app/src
 VOLUME /app/src
 COPY src /app/src
+COPY redis-test-macro /app/redis-test-macro
 
 # Just copy the Cargo.toml files and trigger a build so that we compile our
 # dependencies only. This way we avoid layer cache invalidation if our
--- a/src/checker/http_checker.rs
+++ b/src/checker/http_checker.rs
@@ -119,11 +119,14 @@ fn connection_error(err: &reqwest::Error) -> Option<String> {
     None
 }
 
-fn hyper_error(err: &reqwest::Error) -> Option<String> {
+fn hyper_error(err: &reqwest::Error) -> Option<(CheckStatusReasonType, String)> {
     let mut inner = &err as &dyn Error;
     while let Some(source) = inner.source() {
         if let Some(hyper_error) = source.downcast_ref::<hyper::Error>() {
-            return Some(hyper_error.to_string());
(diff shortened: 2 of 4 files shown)


src/checker/http_checker.rs
```diff
@@ -119,11 +119,17 @@ fn connection_error(err: &reqwest::Error) -> Option<String> {
...
-fn hyper_error(err: &reqwest::Error) -> Option<String> {
+fn hyper_error(err: &reqwest::Error) -> Option<(CheckStatusReasonType, String)> {
     let mut inner = &err as &dyn Error;
     while let Some(source) = inner.source() {
         if let Some(hyper_error) = source.downcast_ref::<hyper::Error>() {
-            return Some(hyper_error.to_string());
+            if hyper_error.is_incomplete_message() {
+                return Some((
+                    CheckStatusReasonType::ConnectionError,
+                    hyper_error.to_string(),
```
wedamija, line 129: Will this string end up being messy, similar to our other errors?

JoshFerge: no it won't be, this will be nice and human readable, as hypers implementation parses their errors.

src/checker/http_checker.rs
```diff
@@ -219,25 +225,29 @@ impl Checker for HttpChecker {
...
-                } else if let Some(message) = hyper_error(&e) {
+                } else if let Some((status_type, message)) = hyper_error(&e) {
                     CheckStatusReason {
-                        status_type: CheckStatusReasonType::Failure,
+                        status_type,
                         description: message,
                     }
                 } else {
+                    // if any error falls through we should log it,
+                    // none should be.
```
evanpurkhiser, line 245:
```suggestion
                    // if any error falls through we should log it,
                    // none should fall through.
```

This is what this meant right?

Dockerfile.localdev
```diff
@@ -30,6 +30,7 @@ WORKDIR /app
 RUN mkdir -p /app/src
 VOLUME /app/src
 COPY src /app/src
+COPY redis-test-macro /app/redis-test-macro
```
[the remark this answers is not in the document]
JoshFerge: yeah

PR #229, opened 2025-01-31
feat(vector): implement spooling logic

JoshFerge:
Right now when a vector request fails, we simply drop the uptime check result. This is not acceptable. The following PR does the following:

- [x] Introduces two new config items `retry_vector_errors_forever`

If `retry_vector_errors_forever` is set to `True`, we will retry making the vector request for a batch forever if the request fails due to a non 2XX or other error. Messages will spool on the channel in the meantime. Given the [current machine size](https://github.com/getsentry/ops/blob/b4fba1ac1e43727a3cd89dddf368ad18da7f6d29/k8s/services/uptime-checker/_values.yaml#L18) and requests per second a pod handles, this should give us roughly 2.5 days of spooling before the service fails from an OOM. (assuming 500 byte check result size, and 4GB pod memory limit).

We also assume for the sake of this idea that we will produce vector results faster than we will uptime checks. I assume this a safe assumption, although if vector is having problems, the queue could still increase while we're producing successfully to vector.

- [x] Introduces a metric for vector request worker queue size, and vector response time.

We can create monitors on both of these metrics, and an aggressive queue size monitor should allow for plenty of time for the on-call to take action before a queue fills up & process dies.

-----------
In a future PR, we could consider a strategy like dropping all _successful_ uptime checks, if the queue is too full. can evaluate options with the team


src/app/config.rs
```diff
@@ -103,6 +103,12 @@ pub struct Config {
     /// The vector endpoint to send results to
     pub vector_endpoint: String,
 
+    /// The maximum number of retries to attempt when sending results to vector
+    pub vector_max_retries: Option<u32>,
+
+    /// Whether to retry sending results to vector indefinitely
+    pub retry_vector_errors_forever: bool,
```
evanpurkhiser, line 110: Seems like we should combine this w/ the other option and make it like `-1` or something maybe?

JoshFerge: i prefer explicit to implicit, and operator error is handled in the case where someone provides an incorrect config

src/producer/vector_producer.rs
```diff
@@ -1,57 +1,113 @@
...
         tracing::info!("vector_worker.starting");
 
         tokio::spawn(async move {
-            let mut batch = Vec::with_capacity(vector_batch_size);
+            let mut batch = Vec::with_capacity(config.vector_batch_size);
 
             while let Some(json) = receiver.recv().await {
+                pending_items.fetch_sub(1, Ordering::SeqCst);
+                metrics::gauge!("producer.pending_items")
+                    .set(pending_items.load(Ordering::SeqCst) as f64);
```
wedamija, line 74: Nice, will be good to be able to keep an eye on this (and alert above some level)

src/app/config.rs
```diff
@@ -112,6 +118,9 @@ pub struct Config {
     /// Allow uptime checks against internal IP addresses
     pub allow_internal_ips: bool,
 
+    /// Whether to disable connection re-use in the http checker
+    pub disable_connection_reuse: bool,
```
wedamija, line 122: Probably need a rebase?

src/app/config.rs
```diff
@@ -173,6 +185,13 @@ impl Config {
         }
 
         let config: Config = builder.extract()?;
+
+        if config.retry_vector_errors_forever && config.vector_max_retries.is_some() {
+            return Err(anyhow::anyhow!(
+                "vector_max_retries must be set to None when retry_vector_errors_forever is true"
+            ));
+        }
```
wedamija, line 193: I think we should warn here in sentry rather than failling. It'd be easy to have a deploy fail because someone forgot to set `vector_max_retries`, but it's not really hurting anything if it's set.

wedamija (commented):
To some extent I wonder if it's even worth having the max number of retries, if we have days of spooling buffer available

src/producer/vector_producer.rs
JoshFerge: yup 100%

src/app/config.rs
JoshFerge: i've removed the `vector_max_retries` option. just have the retry option now

wedamija (approved):

src/producer/vector_producer.rs
```diff
@@ -1,57 +1,108 @@
...
+                    let result = send_batch(
+                        batch_to_send,
+                        client.clone(),
+                        config.endpoint.clone(),
+                        config.retry_vector_errors_forever,
+                    )
+                    .await;
+                    metrics::histogram!("vector_producer.send_batch.duration", "uptime_region" => config.region.clone(), "histogram" => "timer").record(start.elapsed().as_secs_f64());
+                    result
+                } {
```
evanpurkhiser, line 92: nit: Having this whoel block inside the if is getting a bit unweildy, I think we could just do the `if` on the result itself

```rs
if let Err(e) = result {
   // ...
}
```

evanpurkhiser (approved):
Looks good. I think we could clean up some of the metric and logs after in this module to make them all more consistent maybe

klochek (approved):

PR #232, opened 2025-02-05
feat(uptime): Implement redis cluster for config checker

wedamija:
This implements an option to use the redis cluster client for our redis config provider. Defaults to off for backwards compatibility.

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -12,7 +12,9 @@ use tokio_util::sync::CancellationToken;
 use crate::check_executor::{queue_check, CheckSender};
 use crate::config_store::{RwConfigStore, Tick};
 use crate::config_waiter::BootResult;
-use redis::{AsyncCommands, Client};
+use redis::AsyncCommands;
+use crate::redis::build_redis_client;
+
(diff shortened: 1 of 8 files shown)


evanpurkhiser (commented):
I think we just need to update the connection in the scheduler as well?

src/scheduler.rs
```diff
@@ -24,6 +26,7 @@ pub fn run_scheduler(
     redis_host: String,
     config_loaded_receiver: Receiver<BootResult>,
     region: String,
+    redis_enable_cluster: bool,
```
evanpurkhiser, line 28: feels like there might be some smell passing this around, but I think it's fine for now.

evanpurkhiser (approved):

src/scheduler.rs
wedamija: Yeah, feels like maybe we should pass a redis config object around or something?

evanpurkhiser: Yeah or some preconfigured object that can instantiate clients that knows about the config

PR #238, opened 2025-02-06
fix(uptime): fix trace header format

JoshFerge:
we need to remove dashes from the trace_id when we create the trace header. adds assertions on the number of dashes to assure correct format.

### DIFF
--- a/src/checker/http_checker.rs
+++ b/src/checker/http_checker.rs
@@ -173,10 +173,13 @@ fn make_trace_header(config: &CheckConfig, trace_id: &Uuid, span_id: SpanId) -> 
     // according to the service's sampling policy. see
     // https://develop.sentry.dev/sdk/telemetry/traces/#header-sentry-trace
     // for more information.
+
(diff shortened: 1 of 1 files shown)


gaprl (approved):

src/checker/http_checker.rs
```diff
@@ -173,10 +173,13 @@ fn make_trace_header(config: &CheckConfig, trace_id: &Uuid, span_id: SpanId) ->
     // according to the service's sampling policy. see
     // https://develop.sentry.dev/sdk/telemetry/traces/#header-sentry-trace
     // for more information.
+
+    let trace_id_without_dashes = trace_id.to_string().replace("-", "");
```
evanpurkhiser, line 177: I think the better way to do this would be to use the `Simple` formatter https://docs.rs/uuid/latest/uuid/fmt/struct.Simple.html

evanpurkhiser: Could just be `trace_id.simple()` and then the `Display` trait will handle things correctly.

PR #240, opened 2025-02-06
fix(uptime): Fix scheduler to better distribute work amongst the regions

wedamija:
Currently, our checkers in each region are either processing all configs, or are processing none. This pr stripes the work better so that it's evenly distributed between regions

### DIFF
--- a/src/types/check_config.rs
+++ b/src/types/check_config.rs
@@ -23,6 +23,7 @@ pub enum CheckInterval {
 
 /// The largest check interval
 pub const MAX_CHECK_INTERVAL_SECS: usize = CheckInterval::SixtyMinutes as usize;
+pub const SUBSCRIPTION_ID_NAMESPACE: Uuid = Uuid::from_u128(31415926535897932384626433832795u128);
 
 /// The CheckConfig represents a configuration for a single check.
 #[serde_as]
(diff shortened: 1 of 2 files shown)


src/types/check_config.rs
```diff
@@ -23,6 +23,7 @@ pub enum CheckInterval {
 
 /// The largest check interval
 pub const MAX_CHECK_INTERVAL_SECS: usize = CheckInterval::SixtyMinutes as usize;
+pub const SUBSCRIPTION_ID_NAMESPACE: Uuid = Uuid::from_u128(31415926535897932384626433832795u128);
```
evanpurkhiser, line 26: Let's give this a comment since it's magic and arbitrary

src/types/check_config.rs
```diff
@@ -108,9 +111,12 @@ impl CheckConfig {
...
 
-        let running_region_idx = (tick.time().timestamp() / self.interval as i64)
-            .checked_rem(active_regions.len() as i64)
-            .unwrap();
+        let subscription_seed_id =
+            Uuid::new_v5(&SUBSCRIPTION_ID_NAMESPACE, self.subscription_id.as_bytes());
+        let running_region_idx = (((tick.time().timestamp() / self.interval as i64) as u128)
+            + subscription_seed_id.as_u128())
+        .checked_rem(active_regions.len() as u128)
+        .unwrap();
```
evanpurkhiser, line 119: I think it might make sense to have a comment here as well describing why we do this

evanpurkhiser (approved):

PR #250, opened 2025-02-13
fix(uptime): Fix race bug for scheduler progress key

wedamija:
This fixes a bug that occurs when multiple checkers are competing for the same progress key. If a checker saves the key, and then another boots and gets a value too close to the current time then we end up hitting an `OutOfRangeError` when we subtract `start` from `Utc::Now()`.

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -178,13 +178,13 @@ async fn scheduler_loop(
 
 #[cfg(test)]
 mod tests {
-    use std::ops::Add;
     use crate::app::config::Config;
     use crate::check_executor::CheckSender;
     use chrono::{Duration, TimeDelta, Utc};
     use redis::{Client, Commands};
     use redis_test_macro::redis_test;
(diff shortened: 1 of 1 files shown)


evanpurkhiser (approved):

src/scheduler.rs
```diff
@@ -63,7 +63,8 @@ async fn scheduler_loop(
     let start = match result {
         Some(result) => {
             let last_completed_check_nanos: i64 = result.parse().unwrap_or(Utc::now().timestamp());
-            Utc.timestamp_nanos(last_completed_check_nanos) + tick_frequency
+            let next_check = Utc.timestamp_nanos(last_completed_check_nanos) + tick_frequency;
+            next_check.min(Utc::now().duration_trunc(TimeDelta::seconds(1)).unwrap())
```
evanpurkhiser, line 67: Let's add a comment for what this is doing

PR #257, opened 2025-02-14
fix(uptime): Fix slot distribution to use a `uuid5`

wedamija:
Our slot distribution is currently based on the subscription id. The problem with this is that `subscription_id` is already used to partition our configs. This has the side effect of causing configs in a partition to not be distributed evenly among scheduling slots. The result here is that partitions run all at once, then completely stop for a few ticks, then run again, and son on. Using a uuid with a different namespace here will distribute them more evenly.

Some math examples:
Since we modulo subscription ids by 128, then for example, anything that modulos to 51 goes in the same bucket.
For each of those ids in the 51 bucket, if we modulo those by 60 (the number of seconds for an interval), they all map to these numbers
[3, 7, 11, 15, 19, 23, 27, 31, 35, 39, 43, 47, 51, 55, 59]
if i do it for items in partition 50 instead
[2, 6, 10, 14, 18, 22, 26, 30, 34, 38, 42, 46, 50, 54, 58]

Since these all end up in only 1/4 of the total buckets, we end up with spikiness in our checks.

### DIFF
--- a/src/types/check_config.rs
+++ b/src/types/check_config.rs
@@ -24,8 +24,15 @@ pub enum CheckInterval {
 /// The largest check interval
 pub const MAX_CHECK_INTERVAL_SECS: usize = CheckInterval::SixtyMinutes as usize;
 
-/// A random hardcoded id that we use to generated v5 uuids based on subscription ids.
-pub const SUBSCRIPTION_ID_NAMESPACE: Uuid = Uuid::from_u128(31415926535897932384626433832795u128);
+/// A random hardcoded id that we use to generated v5 uuids based on subscription ids for assigning
+/// to slots.
+pub const SUBSCRIPTION_SLOT_NAMESPACE: Uuid =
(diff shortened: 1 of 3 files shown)


src/types/check_config.rs
```diff
@@ -82,11 +89,19 @@ impl CheckConfig {
...
+        // We deterministically produce a new uuid here based on the subscription id and use it to
+        // help distribute subscriptions to slots. Without this, we end up distributing all checks
+        // associated with a specific partition into the same slots.
+        //
+        // We have to use this instead of the subscription id itself because we already use the
+        // subscription id to partition checks.
+        let first_slot = Uuid::new_v5(
+            &SUBSCRIPTION_SLOT_NAMESPACE,
+            self.subscription_id.as_bytes(),
+        )
```
wedamija, line 101: This is the only change here, the rest of this pr is updating tests to choose appropriate uuids

evanpurkhiser (approved):

PR #265, opened 2025-02-19
feat(executor): Add optional retry for check executor

evanpurkhiser:
This adds a new option to allow for retrying failed checks by re-queuing
the check into the executor.

We will continue to record result metrics, but will include a tag
indicating if the result was a retry or not.

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -20,6 +20,8 @@ pub struct ScheduledCheck {
     tick: Tick,
     config: Arc<CheckConfig>,
     resolve_tx: Sender<CheckResult>,
+    /// The number of times this scheduled check has been retried
+    retry_count: u16,
 }
 
 impl ScheduledCheck {
@@ -71,6 +73,7 @@ impl CheckSender {
             tick,
             config,
             resolve_tx,
+            retry_count: 0,
         };
 
         self.queue_size.fetch_add(1, Ordering::SeqCst);
@@ -79,6 +82,15 @@ impl CheckSender {
             .expect("Failed to queue ScheduledCheck");
 
         resolve_rx
+    }
+
+    /// Requeues the check to be executed again, increasing the number of retries by 1
+    fn queue_check_for_retry(&self, mut check: ScheduledCheck) {
+        check.retry_count += 1;
+        self.queue_size.fetch_add(1, Ordering::SeqCst);
+        self.sender
+            .send(check)
+            .expect("Failed to queue ScheduledCheck");
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/check_executor.rs
```diff
@@ -103,55 +115,78 @@ impl CheckResult {
...
     tracing::info!("executor.starting");
 
-    let (check_sender, receiver) = CheckSender::new();
-    let queue_size = check_sender.queue_size.clone();
-    let num_running = check_sender.num_running.clone();
+    let (sender, reciever) = CheckSender::new();
+    let queue_size = sender.queue_size.clone();
+    let num_running = sender.num_running.clone();
+
+    let check_sender = Arc::new(sender);
```
evanpurkhiser, line 129: the sender now needs to be owned by the executor AND the external service sending checks, so we put it into an arc.

src/check_executor.rs
```diff
@@ -103,55 +115,78 @@ impl CheckResult {
...
+    let executor_check_sender = check_sender.clone();
 
-    let executor = tokio::spawn(async move {
+    let conf = ExecutorConfig {
+        concurrency,
+        failure_retries,
+        region,
+    };
```
evanpurkhiser, line 136: small cleanup, clippy was complaining that we're giving the `executor_loop` function too many arguments, so just moved some into a config type thing

src/check_executor.rs
```diff
@@ -80,6 +83,15 @@ impl CheckSender {
...
     }
+
+    /// Requeues the check to be executed again, increasing the number of retries by 1
+    fn queue_check_for_retry(&self, mut check: ScheduledCheck) {
+        check.retry_count += 1;
+        self.queue_size.fetch_add(1, Ordering::SeqCst);
+        self.sender
+            .send(check)
+            .expect("Failed to queue ScheduledCheck");
+    }
```
wedamija, line 94: At some point we might want to add some kind of wrapper around this queue so that we're not duplicating code

src/check_executor.rs
```diff
@@ -170,11 +205,27 @@ async fn executor_loop(
...
+                        &check_result,
+                        scheduled_check.retry_count > 0,
+                        will_retry,
+                    );
+
+                    // re-queue for execution again
+                    if will_retry {
+                        tracing::debug!(result = ?check_result, "executor.check_will_retry");
+                        job_check_sender.queue_check_for_retry(scheduled_check);
+                        job_num_running.fetch_sub(1, Ordering::SeqCst);
```
wedamija, line 221: Let's move this out of the if statement, and just put it before `will_retry` so that we're not duplicating?

src/check_executor.rs
```diff
@@ -170,11 +205,27 @@ async fn executor_loop(
...
+                    }
+
                     if let Err(e) = job_producer.produce_checker_result(&check_result) {
                         tracing::error!(error = ?e, "executor.failed_to_produce");
                     }
 
-                    record_result_metrics(&check_result);
                     tracing::debug!(result = ?check_result, "executor.check_complete");
 
                     scheduled_check.record_result(check_result);
```
wedamija, line 231: Is there any problem with delaying the recording of the result here? I forget if anything is waiting on it

wedamija (approved):

src/check_executor.rs
evanpurkhiser: I think this is OK actually, where it's located _is_ the wrapper

src/check_executor.rs
evanpurkhiser: 👍

evanpurkhiser: Actally I'm going to leave this here, just since we also do the sending with the executor. so if we move it up we'll be skipping the sending bit and the metric won't be quite accurate

PR #267, opened 2025-02-24
feat(uptime): add redirect errors

JoshFerge:
note that we can't deploy this untill a corresponding `redirect_error` has been added to the sentry kafka repo and then the uptime checker, sentry, and snuba have all been bumped with the new version of the library


wedamija (approved):
lgtm, but we should have a test here

JoshFerge:
> lgtm, but we should have a test here

test added

PR #268, opened 2025-02-24
feat: catch hyper util errors

JoshFerge:
looking at this https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0A-jsonPayload.message%3D%22scheduler.tick_scheduled%22%0Alabels.%22k8s-pod%2Fservice%22%3D%22uptime-checker%22%0A-%22DNS%22;summaryFields=:true:32:beginning;cursorTimestamp=2025-02-24T22:52:57.665242903Z;duration=PT15S?project=internal-sentry

we can see there are a lot of raw errors making it through. some of them are these hyper_util errors. we need to pin the library from being a sub-dependency so we can use the error here to cast it and return the parsed error.

it looks like there are some ssl errors which can make it into this category, so can follow up to try to separate those. specifically 
```
check_url.error: reqwest::Error { kind: Request, source: hyper_util::client::legacy::Error(Connect, Ssl(Error { code: ErrorCode(5), cause: None }, X509VerifyResult { code: 0, error: "ok" })) }
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4270,6 +4270,7 @@ dependencies = [
  "hostname 0.4.0",
  "httpmock",
  "hyper 1.4.1",
+ "hyper-util",
  "ipnet",
  "metrics",
  "metrics-exporter-statsd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ rust-version = "1.80"
 
 [dependencies]
 hyper = "1.4.1"
+hyper-util = "0.1.10"
 anyhow = "1.0.66"
 clap = { version = "4.4.6", features = ["derive"] }
 chrono = { version = "0.4.31", default-features = false, features = ["std", "serde"] }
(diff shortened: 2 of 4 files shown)


wedamija (approved):
lgtm, although we should add a test

JoshFerge:
> lgtm, although we should add a test

will see if i can test this functionality -- have had mixed results trying to replicate to the right connection errors.

JoshFerge:
@wedamija I was unable to create a test which causes a hyper util legacy error to throw. I feel confident based on the type checking that this will work -- what do you think?

wedamija:
> @wedamija I was unable to create a test which causes a hyper util legacy error to throw. I feel confident based on the type checking that this will work -- what do you think?
I think that's ok... it's also specific to the library, so if we change libraries it won't be relevant, so we don't need a test to save us there either

wedamija (approved):

PR #271, opened 2025-02-26
feat(metrics): Monitor executor task metrics

evanpurkhiser:
This should help us to understand if our http checker tasks are taking a suspiciously long time to be scheduled or polled.

Our suspicion is that when we turn up execution concurrency we may be overloading the tokio scheduler in a way such that we are very slow to poll the http tasks. By the time we finally poll the task the reqwest timeout may have been reached and the request will be considered timed out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1536,7 +1536,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.10",
+ "socket2 0.5.7",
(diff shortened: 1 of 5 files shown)


evanpurkhiser:
Once merged we're going to want to follow this guide https://docs.rs/tokio-metrics/latest/tokio_metrics/struct.TaskMonitor.html#why-are-my-tasks-slow

wedamija (approved):

PR #280, opened 2025-03-07
feat(uptime): Allow dns nameservers to be specified for the http checker

wedamija:
This allows us to point our http checker at a different dns server. This should allow us to circumvent kube-dns for our checks, which will hopefully resolve a lot of dns issues we've been seeing.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -1,5 +1,5 @@
 use std::{borrow::Cow, collections::BTreeMap, net::SocketAddr};
-
+use std::net::IpAddr;
 use figment::{
     providers::{Env, Format, Serialized, Yaml},
     Figment,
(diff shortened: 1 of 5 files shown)


evanpurkhiser (approved):

src/app/config.rs
```diff
@@ -314,6 +320,7 @@ mod tests {
                 ("UPTIME_CHECKER_CHECKER_NUMBER", "2"),
                 ("UPTIME_CHECKER_TOTAL_CHECKERS", "5"),
                 ("UPTIME_CHECKER_FAILURE_RETRIES", "2"),
+                ("UPTIME_CHECKER_HTTP_CHECKER_DNS_NAMESERVERS", "8.8.8.8,8.8.4.4")
```
JoshFerge, line 323: do we know if google's public DNS servers do any kind of rate limiting?

wedamija: Not sure, This is just a test that is reading the values though

PR #295, opened 2025-04-11
fix(uptime): Have uptime-checker shut down gracefully on sigterm

klochek:
The executor will block indefinitely until all feeding producers are dropped.  We weren't shutting down the partitioned services when a shutdown was issued, but even with fixing that, the executor still holds on to a producer that it uses for its own retries.  So, just add a `take_until` on the stream, awaiting a future that is triggered by the cancel signal.

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -397,8 +397,10 @@ mod tests {
 
     use chrono::Utc;
     use futures::poll;
+    use ntest::timeout;
     use similar_asserts::assert_eq;
     use tokio::{sync::oneshot::Receiver, time};
+    use tokio_util::sync::CancellationToken;
     use uuid::Uuid;
 
     use super::*;
@@ -415,6 +417,9 @@ mod tests {
             status: CheckStatus::Success,
         };
 
+        let cancel_token = CancellationToken::new();
+        let cancel_future = cancel_token.cancelled_owned();
+
         let dummy_checker = DummyChecker::new();
         dummy_checker.queue_result(delayed_result);
 
@@ -427,7 +432,7 @@ mod tests {
             region: "us-west".to_string(),
             record_task_metrics: false,
         };
-        let (sender, _) = run_executor(checker, producer, conf);
+        let (sender, _) = run_executor(checker, producer, conf, cancel_future);
 
         let tick = Tick::from_time(Utc::now());
         let config = Arc::new(CheckConfig {
@@ -454,6 +459,9 @@ mod tests {
             status: CheckStatus::Success,
         };
 
+        let cancel_token = CancellationToken::new();
+        let cancel_future = cancel_token.cancelled_owned();
+
         let dummy_checker = DummyChecker::new();
         dummy_checker.queue_result(delayed_result.clone());
         dummy_checker.queue_result(delayed_result.clone());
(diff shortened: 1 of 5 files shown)


src/check_executor.rs
```diff
@@ -136,6 +137,7 @@ pub fn run_executor(
     checker: Arc<impl Checker + 'static>,
     producer: Arc<impl ResultsProducer + 'static>,
     conf: ExecutorConfig,
+    cancel_future: WaitForCancellationFutureOwned,
```
evanpurkhiser, line 140: This is sort of different from a [`CancelationToken`](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html) right?

src/check_executor.rs
```diff
@@ -193,6 +197,7 @@ async fn executor_loop(
     }
 
     schedule_check_stream
+        .take_until(cancel_future)
```
evanpurkhiser, line 200: super clean

src/check_executor.rs
```diff
@@ -392,8 +397,10 @@ mod tests {
 
     use chrono::Utc;
     use futures::poll;
+    use ntest::timeout;
```
evanpurkhiser, line 400: What's `ntest` for?

src/manager.rs
```diff
@@ -206,10 +211,11 @@ impl Manager {
 
                 results_worker.await.expect("Failed to stop vector worker");
 
-                executor_join_handle.await.expect("Failed to stop executor");
                 services_join_handle
                     .await
                     .expect("Failed to stop partitioned services");
+
+                executor_join_handle.await.expect("Failed to stop executor");
```
evanpurkhiser, line 218: does it have to move after like this?

src/manager.rs
```diff
@@ -161,7 +165,8 @@ impl Manager {
                 ));
                 // XXX: Executor will shutdown once the sender goes out of scope. This will happen once all
                 // referneces of the Sender (executor_sender) are dropped.
-                let (sender, handle) = run_executor(checker.clone(), producer, executor_conf);
+                let (sender, handle) =
+                    run_executor(checker.clone(), producer, executor_conf, cancel_future);
```
evanpurkhiser, line 169: would it actually make sense for this to accept a cloned `CancelationToken` instead of the `WaitForCancellationFutureOwned`, can we just get one of those right where we need it in the take?

src/check_executor.rs
klochek: It's a future that's awaiting that cancellation token

src/check_executor.rs
klochek: That's where the test timeout macro comes from

src/check_executor.rs
evanpurkhiser: Ah yeah I see

src/manager.rs
klochek: Strictly speaking, no, this should still work in the original order, but I think conceptually it makes more sense to await shutdown of the producers first before finally ending the executor.  (This is also what I did before I realized I couldn't cleanly take down the executor just by stopping the producers, due to the executor's own hold on the channel.)

src/manager.rs
klochek: Good idea!  Cleaner that way.

evanpurkhiser (approved):
nice!

PR #296, opened 2025-04-22
gha: pin versions and set permissions

mdtro:
1. Using version tags or no specifier at all can open us up to dependency attacks.
    Pinning the SHA is a strong mitigation. 
    See: https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/
    
2. Set [`permissions`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token) on the jobs so they are restricted to only what is necessary.

3. Where appropriate, use `env:` to [avoid potential shell injection attacks](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#good-practices-for-mitigating-script-injection-attacks). 

4. Fix docker build command syntax.


evanpurkhiser (approved):

.github/workflows/build-image.yaml
```diff
@@ -16,48 +17,69 @@ jobs:
...
+              --tag "${VERSIONED_IMAGE}" \
+              --file "${DOCKERFILE_PATH}" \
+              --push \
+              .
+          else
+            docker buildx build \
+              --platform "${PLATFORMS}" \
+              --tag "${VERSIONED_IMAGE}" \
+              --file "${DOCKERFILE_PATH}" \
+              --push \
```
aldy505, line 55: AFAIK, `--push` should only happen on main branch. But please confirm with the devinfra team.

mdtro: Ah good catch!

mdtro:
I ended up reverting https://github.com/getsentry/uptime-checker/pull/158 to unblock CI.
@aldy505 I can work with you tomorrow on building out these workflows if you'd like. I have some bugs in my own. 😅

aldy505:
We basically hit this https://github.com/docker/buildx/issues/59

mdtro:
Closing in favor of: https://github.com/getsentry/uptime-checker/pull/303

PR #300, opened 2025-04-23
ci: use docker action instead of using the docker commands directly

aldy505:


aldy505:
cc @mdtro

aldy505:
Closed due to https://github.com/getsentry/uptime-checker/pull/303

PR #302, opened 2025-04-25
fix(uptime): Add timeouts for redis rpcs; process many tick results at once

klochek:
Add a configurable (default 30s) connection/response timeouts, so that we don't block indefinitely awaiting a response that never comes.  For the scheduler specifically, break out and log the individual connection errors we might see, so that we just log timeouts, but retry connecting if there's an actual connection drop.  

Also in the scheduler, process many tick results at once using recv_many, and then simply use the last one in the (ordered) vec of results as the appropriate progress key.

Finally, update redis to 0.30.0.


src/redis.rs
```diff
@@ -58,20 +59,34 @@ pub fn build_redis_client(redis_url: &str, enable_cluster: bool) -> RedisClient
...
-            RedisClient::Single(client) => RedisAsyncConnection::Single(
-                client
-                    .get_multiplexed_tokio_connection()
-                    .await
-                    .expect("Unable to connect to Redis"),
-            ),
+            RedisClient::Cluster(client) => {
+                let mut config = ClusterConfig::new();
+
+                if redis_timeouts_ms > 0 {
```
evanpurkhiser, line 71: Do we actually need this condtion? when would we end up setting this to 0?

src/scheduler.rs
```diff
@@ -159,21 +165,56 @@ async fn scheduler_loop(
     };
 
     let in_order_tick_processor_join_handle = tokio::spawn(async move {
-        while let Some(tick_complete_join_handle) = tick_complete_rx.recv().await {
+        const HANDLE_QUEUE_SIZE: usize = 1000;
+        let mut tick_handles = Vec::with_capacity(HANDLE_QUEUE_SIZE);
```
evanpurkhiser, line 169: What happens if it reaches capacity?

evanpurkhiser: Oh sorry misreading

src/scheduler.rs
```diff
@@ -159,21 +165,56 @@ async fn scheduler_loop(
     };
 
     let in_order_tick_processor_join_handle = tokio::spawn(async move {
-        while let Some(tick_complete_join_handle) = tick_complete_rx.recv().await {
+        const HANDLE_QUEUE_SIZE: usize = 1000;
```
evanpurkhiser, line 168: I think this would be better named `HANDLE_BATCH_SIZE`

src/redis.rs
klochek: I should probably add a comment here: local tokio tests use paused time, which causes all timeouts to trigger immediately, failing the test.

src/scheduler.rs
```diff
@@ -159,21 +165,56 @@ async fn scheduler_loop(
...
+            let num_read = tick_handles.len();
+
+            // Only read the very last update's tick, since that will be the latest one we've successfully
+            // processed.
+            let tick = tick_handles.pop();
+            tick_handles.clear();
+
+            let tick = tick
+                .unwrap()
                 .await
```
wedamija, line 188: I think that we need to await all the ticks in the batch, even if we just store the last one. Otherwise, we might commit before we're sure everything has run.

src/scheduler.rs
```diff
@@ -159,21 +165,56 @@ async fn scheduler_loop(
...
+
+                    connection = client
+                        .get_async_connection(redis_timeouts_ms)
+                        .await
+                        .expect("Unable to connect to redis");
+                } else {
+                    tracing::error!(progress_key, error = %e, "scheduler.progress_fatal_error");
+
+                    panic!("Could not save scheduler progress: {:?}", e);
+                }
```
wedamija, line 216: Could we add some tests that cover this code?

wedamija (commented):
It might be a good idea to split the timeout change and the multiple tick results into separate prs, just so you can get one part merged first

src/scheduler.rs
klochek: I'm actually a little confused on the semantics of this--if some check within a tick returns an error, we'll panic in the tick_complete_join_handle task, which in turn will panic in the in_order_tick_processor_join_handle task, and so this particular scheduler will no longer write to progress to redis (but it will still keep executing checks).  How do we get it started again?

klochek:
> It might be a good idea to split the timeout change and the multiple tick results into separate prs, just so you can get one part merged first

Only thing that concerns me here is that with retries, we could end up in a situation were we have to clear 30s worth of a backlog, which will add pressure to the scheduler.

src/scheduler.rs
klochek: Not easily, no :( I tried doing this with mocks (which seems like the "correct" approach to me,) but trying to get multiple levels of nested mocks working seems impossible with current frameworks (I wanted to mock the RedisAsyncConnection returned by our client, so that I could have it return a timeout error, which means mocking our client and other bits, in order to return a mocked connection struct.)

I also tried this more directly by using the fact that paused tokio tests will automatically complete time-related futures: create a connection with a non-zero timeout, and start the test unpaused.  Once we're past initialization, pause the test, and let the scheduler run against some connections that ought to immediately timeout.  In reality, this seems to be less consistent than implied by the docs, and I can see some redis calls are intermittantly succeeding without timing out (and it happens constantly in CI.)

Any ideas?

src/scheduler.rs
wedamija: What cases would cause it to panic? Probably, if a check fails we shouldn't cause this to die.

Maybe this is what caused us to lose progress that time? We panicked, killed the progress thread and then at some point restarted?

wedamija (approved):

src/scheduler.rs
klochek: Okay, this is important :). I will commit this PR as is, but am going to follow up with something that is less panicky than our current error handling.

PR #304, opened 2025-05-05
ref(uptime): rationalize usage of unwrap/expect; prefer Result<> to expect in some cases

klochek:
1. `expect` ought to be called with the reason we think the preceding call will not fail; update our usages accordingly
2. Prefer `expect` over `unwrap`, as the additional context/reasoning of `expect` is helpful
3. Remove some `expects` from lower-level code, replacing them with `Result<...>`s that propagates up the stack 
4. Replace some `expects` around things that shouldn't trigger complete failures with matching and tracing (in the scheduler tasks in particular)

### DIFF
--- a/src/check_config_provider/redis_config_provider.rs
+++ b/src/check_config_provider/redis_config_provider.rs
@@ -1,20 +1,16 @@
-use redis::AsyncCommands;
-use std::collections::HashSet;
-use std::sync::Arc;
-use tokio::task::JoinHandle;
-use tokio_util::sync::CancellationToken;
-
+use crate::redis::RedisClient;
 use crate::{app::config::Config, manager::Manager, types::check_config::CheckConfig};
-
-use chrono::Utc;
+use anyhow::Result;
+use redis::AsyncCommands;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+use std::collections::HashSet;
+use std::sync::Arc;
 use std::time::Duration;
+use tokio::task::JoinHandle;
 use tokio::time::interval;
+use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
-
-#[allow(unused_imports)]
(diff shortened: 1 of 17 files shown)


src/check_config_provider/redis_config_provider.rs
```diff
@@ -1,21 +1,17 @@
...
+use redis::AsyncCommands;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+use std::collections::HashSet;
+use std::sync::Arc;
 use std::time::Duration;
+use tokio::task::JoinHandle;
 use tokio::time::interval;
+use tokio_util::sync::CancellationToken;
 use uuid::Uuid;
```
evanpurkhiser, line 13: Is there something re-sorting all these?

src/check_config_provider/redis_config_provider.rs
```diff
@@ -153,39 +149,35 @@ impl RedisConfigProvider {
...
-                .unwrap()
-                .as_secs_f64();
             metrics::histogram!(
                 "config_provider.initial_load.partition.duration",
                 "histogram" => "timer",
                 "uptime_region" => region.clone(),
                 "partition" => partition.number.to_string(),
             )
-            .record(partition_loading_time);
+            .record(partition_start_loading.elapsed().as_secs_f64());
```
evanpurkhiser, line 172: minor nit, I sort of prefer to keep the variable name here since it helps document what this value is, vs having to parse this expression and understand what it is

evanpurkhiser:
Any chance you can split out all the `expect` text updates into their own PR? Happy to approve that right away. Need to look closer at some of the other ones.

klochek:
> Any chance you can split out all the `expect` text updates into their own PR? Happy to approve that right away. Need to look closer at some of the other ones.

Sure thing, over here: https://github.com/getsentry/uptime-checker/pull/305

PR #305, opened 2025-05-05
ref(uptime): clean up unwrap/expect usage

klochek:
1. `expect` ought to be called with the reason we think the preceding call will not fail; update our usages accordingly
2. Prefer `expect` over `unwrap`, as the additional context/reasoning of `expect` is helpful

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,7 +14,7 @@ use crate::{
 
 pub fn execute() -> io::Result<()> {
     let app = cli::CliApp::parse();
-    let config = Arc::new(config::Config::extract(&app).expect("Configuration invalid"));
+    let config = Arc::new(config::Config::extract(&app).expect("Config should be valid"));
 
     logging::init(LoggingConfig::from_config(&config));
     metrics::init(&config.metrics);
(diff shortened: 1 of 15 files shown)


klochek:
> Oh sorry maybe I didn't clarify, I meant for this PR to be ONLY the string `expect` changes. Like these
> 
> ```diff
> -    let config = Arc::new(config::Config::extract(&app).expect("Configuration invalid"));
> +    let config = Arc::new(config::Config::extract(&app).expect("Config should be valid"));
> ```
> 
> since these are very easy to review. The other changes I just wanted to look closer at.

Ah, sorry--updated!

evanpurkhiser:
Looks like there are still some `unwrap` → `expects`. Sorry it was just stuff like this I wanted to look closer at while not having to see all the text only changes 😅

https://github.com/getsentry/uptime-checker/pull/305/files#diff-3d8724b3c76e2e43e07e0dcc66176a94d7735fc7736cd11bc8f4b00da454a164R101

klochek:
> Looks like there are still some `unwrap` → `expects`. Sorry it was just stuff like this I wanted to look closer at while not having to see all the text only changes 😅
> 
> https://github.com/getsentry/uptime-checker/pull/305/files#diff-3d8724b3c76e2e43e07e0dcc66176a94d7735fc7736cd11bc8f4b00da454a164R101

But, an unwrap _is_ an expect, just minus the helpful text; the semantics around it have not changed....

evanpurkhiser:
Isn't the `unwrap_or_default` a possible behavior change though?

evanpurkhiser (approved):

evanpurkhiser:
> But, an unwrap is an expect, just minus the helpful text; the semantics around it have not changed....

Yeah I know, the way my brain ends up reviewing things is that I really like when there's a specific class of changes that can be classified as 'trivial' that I can approve immediately, (like just changing some expect strings), vs introducing new expect strings I would want to take a bit closer look at what the expect strings are.

I know I can do both of these in one review, and looking at the two types of changes are pretty similar (they're both reviewing expect strings), but my brain just likes them separate haha

klochek:
> Isn't the `unwrap_or_default` a possible behavior change though?

My eyes are going--I didn't catch that one.

PR #307, opened 2025-05-07
ref(uptime): return Result<> structs and log/trace instead of panicking in the scheduler loop

klochek:
Part 1: scheduler loop.  I think this is the sane pattern we can repeat in our code for our other tasks: no more panicking, return errors, and the top-level function that is launched inside a tokio::spawn gets access to the common `tasks_finished` channel that will receive the return result from that function.  That in turn goes to the receiver in the app, which can then make a decision about what (if anything) to do with the result.  In this PR, if there's any kind of error, we'll just panic and die, but in the future we could put in some selective retries.

This does not address all the looping tasks we have; we still have the redis writer and the executor.  Those can be dealt with in forthcoming PRs.  If a panic does occur in any of those tasks, the resulting channels they use to communicate with are closed, and subsequent reads/writes to them generate errors in the scheduler, which lead to the entire service being brought down--it's just very circuitous and requires judicious log tracing to understand why.

### DIFF
--- a/src/scheduler.rs
+++ b/src/scheduler.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use chrono::{DurationRound, TimeDelta, TimeZone, Utc};
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


src/scheduler.rs
```diff
@@ -57,16 +64,10 @@ async fn scheduler_loop(
...
-        .await
-        .expect("Unable to connect to redis");
-    let last_progress: Option<String> = connection
-        .get(&progress_key)
-        .await
-        .expect("Unable to get last tick key");
+) -> anyhow::Result<()> {
+    let client = build_redis_client(&redis_host, redis_enable_cluster)?;
+    let mut connection = client.get_async_connection(redis_timeouts_ms).await?;
+    let last_progress: Option<String> = connection.get(&progress_key).await?;
```
wedamija, line 70: If these error, is it possible to annotate them with what happened? I think it'd be more helpful to know that redis failed when we attempted to get the key, for example. I guess we'll also know that from the stack trace though

wedamija (approved):
lgtm. Since this is a big change, could you also test running this locally and make sure it's all working?

We have good tests for each model, but don't have a full integration test yet

PR #308, opened 2025-05-08
fix(uptime): Add retry count to check trace id

klochek:


src/check_executor.rs
```diff
@@ -228,7 +228,9 @@ async fn executor_loop(
                     let check_result = if late_by > interval {
                         CheckResult::missed_from(config, tick, &job_region)
                     } else {
-                        job_checker.check_url(config, tick, &job_region).await
+                        job_checker
+                            .check_url(config, tick, &job_region, scheduled_check.retry_count)
```
evanpurkhiser, line 232: I'm not in love with drilling this in, and almost wonder if we shuld pass the ScheduledCheck instead of the Tick instead

evanpurkhiser (approved):

src/check_executor.rs
klochek: Yeah, I think you're right, it looks better.  Updated.

PR #310, opened 2025-05-15
ref(uptime): factor check_executor_loop

klochek:
In preparation for more changes; plus, it's a little easier to reason about.  Also, always instrument the resulting future.  (This is actually vital, as we otherwise start running into awful type problems with differently-typed futures being returned from the, but also, it's what we run in prod.)

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -226,18 +226,26 @@ async fn executor_loop(
             let job_check_sender = check_sender.clone();
             let job_queue_size = queue_size.clone();
 
-            let check_fut = do_check(
-                &conf,
-                scheduled_check,
-                job_queue_size,
-                job_num_running,
-                job_checker,
-                job_check_sender,
-                job_producer,
-                job_region,
-            );
-
-            metrics_monitor.instrument(check_fut)
+            async {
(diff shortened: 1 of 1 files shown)


src/check_executor.rs
```diff
@@ -224,81 +224,88 @@ async fn executor_loop(
...
-                };
-
-                if conf.record_task_metrics {
-                    job_metrics_monitor.instrument(check_task).await;
-                } else {
-                    // Expect is necessary here--we need to break out of the stream processing loop.
-                    tokio::spawn(check_task)
-                        .await
-                        .expect("The check task should not fail");
-                }
```
evanpurkhiser, line 295: Are we getting rid of this config?

evanpurkhiser (approved):

src/check_executor.rs
klochek: bleh, turns out we don't enable tokio recording in de, s4s, anywhere but us.  I've managed to restore the original logic.)

evanpurkhiser: I'm OK just killing this setting also now hah

wedamija (approved):

PR #311, opened 2025-05-20
ref(uptime): use and cache gauge handles

klochek:
Helps reduce some allocations

### DIFF
--- a/src/check_executor.rs
+++ b/src/check_executor.rs
@@ -216,23 +216,29 @@ async fn executor_loop(
         });
     }
 
+    let queue_gauge =
+        metrics::gauge!("executor.queue_size", "uptime_region" => conf.region.clone());
+    let num_running_gauge =
+        metrics::gauge!("executor.num_running", "uptime_region" => conf.region.clone());
+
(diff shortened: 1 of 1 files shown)


src/check_executor.rs
```diff
@@ -216,23 +216,29 @@ async fn executor_loop(
...
             let job_region = conf.region.clone();
-            let job_num_running = num_running.clone();
             let job_check_sender = check_sender.clone();
-            let job_queue_size = queue_size.clone();
+
+            let queue_size_val = queue_size.fetch_sub(1, Ordering::Relaxed) - 1;
+            queue_gauge.set(queue_size_val as f64);
+
+            let num_running_val = num_running.fetch_add(1, Ordering::Relaxed) + 1;
+            num_running_gauge.set(num_running_val as f64);
```
wedamija, line 236: Why do we do `-1` and `+1` here? Aren't the calls already doing that?

wedamija: Makes sense, it's even in the name!

wedamija (approved):

PR #312, opened 2025-05-21
ref(uptime): leak region string to avoid many allocations

klochek:
At the cost of some unfreeable memory (which won't be freed until the program shuts down, anyway) we avoid making string clones/heap allocations, especially in the inner scheduler and executor loops.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -111,7 +111,8 @@ pub struct Config {
     pub redis_host: String,
 
     /// The region that this checker is running in
-    pub region: String,
+    #[serde(default, deserialize_with = "deserialize_region")]
+    pub region: &'static str,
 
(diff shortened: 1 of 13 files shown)


wedamija (approved):
This makes sense to me - these strings never change throughout execution.

What kind of performance improvements do changes like this make?

klochek:
> What kind of performance improvements do changes like this make?

These are generally small (2-3% improvement), but they have the cumulative effect of making profiles really noisy-looking and harder to read.  There's also an issue around tracing/metric calls, which are much more efficient with borrowed strs (computing a key is just hashing a pointer, instead of doing a strcmp.). As I've been doing this work, the fraction of work attributable to our tracing instrumentation keeps growing....

PR #317, opened 2025-05-29
feat(uptime): allow checker to run checks in parallel (via config)

klochek:


src/check_executor.rs
```diff
@@ -241,7 +244,13 @@ async fn executor_loop(
                     conf.region,
                 );
                 if conf.record_task_metrics {
-                    metrics_monitor.instrument(check_fut).await;
+                    if conf.checker_parallel {
+                        tokio::spawn(metrics_monitor.instrument(check_fut))
+                            .await
+                            .expect("The check task should not fail");
```
evanpurkhiser, line 250: I actually don't think we need an option for this. It was enabled before just fine before we started collecting metrics 😅

klochek: I'd really love to compare between the two modes in de, though.  Once that work is complete, I'm happy to get rid of all these configs and just do the spawn-instrumented....

evanpurkhiser: sounds good!

evanpurkhiser (approved):

PR #325, opened 2025-06-27
chore: bump Rust toolchain version to 1.85

lcian:
Bumps the Rust version to 1.85 and fixes the new lints.
Needed for #324


evanpurkhiser:
Going to merge since CI seems broken right now

PR #327, opened 2025-06-30
ci: Add repo-token to setup-protoc

evanpurkhiser:


Jeffreyhung (approved):
interesting that they don't just grab the token automatically https://github.com/arduino/setup-protoc

PR #350, opened 2025-07-02
ci: ammend manifest to combine multi-platform image

mdtro:
I _believe_ this should work to ammend the SHA and nightly manifest as multi-platform.

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -60,6 +60,10 @@ jobs:
           - os: ubuntu-24.04-arm
             platform: arm64
 
+    outputs:
+      digest: ${{ steps.build_push.outputs.digest }}
+      metadata: ${{ steps.meta.outputs.json }}
+
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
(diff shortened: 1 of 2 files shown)


.github/workflows/image.yml
```diff
@@ -90,6 +90,7 @@ jobs:
           tags: |
             type=sha,prefix=
             type=sha,prefix=,format=long
+            type=sha,prefix=,format=long,suffix=-${{ matrix.platform }}
```
evanpurkhiser, line 93: Should we even push these other's then? seems like w should JUST push the sha w/ platform?

.github/workflows/image.yml
```diff
@@ -112,3 +111,40 @@ jobs:
...
+        with:
+          username: sentrybuilder
+          password: ${{ secrets.DOCKER_HUB_RW_TOKEN }}
+
+      - name: Create and push multi-platform manifest
+        run: |
+          ./scripts/create-manifest.sh \
+            "${{ github.repository }}" \
+            "${{ github.sha }}" \
+            "${{ github.ref == format('refs/heads/{0}', github.event.repository.default_branch) }}"
```
evanpurkhiser, line 149: What does this end up evaluating to? `1/0` `true/false`?

evanpurkhiser (approved):

.github/workflows/image.yml
mdtro: I think `true`? 
https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions#operators

.github/workflows/image.yml
```diff
@@ -112,3 +111,40 @@ jobs:
...
           sbom: true
           push: true
+
+  create-manifest:
+    name: Create multi-platform manifest
+    needs: build-main-or-release
+    if: |
+      github.repository_owner == 'getsentry'
+      && github.event_name == 'push'
+      && (github.ref == format('refs/heads/{0}', github.event.repository.default_branch) || startsWith(github.ref, 'refs/heads/release/'))
```
evanpurkhiser, line 121: Do we actually even need this condition, since it already needs `build-main-or-release`

PR #354, opened 2025-07-03
ci: craft should wait for "Create multi-platform manifest" job to finish

aldy505:


aldy505:
@wedamija @evanpurkhiser this is due in 5 days. Can this get merged?

PR #362, opened 2025-07-16
ci: run self-hosted e2e test

aldy505:
I am trying to add this e2e test across repositories that self-hosted depends on. Hopefully it'll make tracking broken commits easier in the future.

PR #373, opened 2025-07-22
ci: run self-hosted e2e test (second attempt)

aldy505:
The failure from https://github.com/getsentry/uptime-checker/pull/362 was simply caused by wrong `project_name`. It should be `uptime_checker` rather than `uptime-checker`.


aldy505:
cc @evanpurkhiser

evanpurkhiser (approved):

PR #376, opened 2025-07-22
feat(uptime): add redirect uris and connection start timestamps to stats

klochek:


wedamija (approved):

evanpurkhiser:
@klochek I'm reverting this since I'm trying to get master into a state where it reproduces the exact build we had for when we were able to deploy something that works

PR #391, opened 2025-08-06
feat(uptime): add CertificateInfo to CheckResult

klochek:
Just expose not_before/not_after timestamps

### DIFF
--- a/src/types/result.rs
+++ b/src/types/result.rs
@@ -322,7 +322,8 @@ mod tests {
         "start_us": 5,
         "duration_us": 5
       }
-    }
+    },
+    "certificate_info": null
   },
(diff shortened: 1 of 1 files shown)


src/types/result.rs
```diff
@@ -162,6 +182,10 @@ pub struct RequestInfo {
     pub request_duration_us: u64,
 
     pub durations: RequestDurations,
+
+    // Information about the leaf certificate retrieved as part of this
+    // request, if any.
```
evanpurkhiser, line 187: I wonder if at some point we would want to consider interim certs as well

evanpurkhiser (approved):

PR #395, opened 2025-10-06
feat(uptime): check robots.txt once per day, per subscription

klochek:


src/checker.rs
```diff
@@ -49,18 +51,36 @@ pub trait Checker: Send + Sync {
...
         check: &ScheduledCheck,
         region: &'static str,
+        check_robots: bool,
     ) -> impl Future<Output = CheckResult> + Send;
 }
 
 #[derive(Debug)]
 pub enum HttpChecker {
     ReqwestChecker(ReqwestChecker),
+    DummyChecker(DummyChecker),
```
evanpurkhiser, line 61: ah I see

src/checker/reqwest_checker.rs
```diff
@@ -229,13 +230,58 @@ impl Checker for ReqwestChecker {
...
+                        res.bytes().await.unwrap_or_default()
+                    } else {
+                        tracing::debug!("could not retrieve robots.txt");
+                        Bytes::default()
+                    }
+                };
+
+                if let Ok(r) = Robot::new("SentryUptimeBot", &robots_txt) {
+                    if !r.allowed(url.as_str()) {
+                        return CheckResult {
```
evanpurkhiser, line 261: Hmm, doesn't this mean we're actually skipping a normal check then though? So when we check their robots once a day we won't have checked their configured URL and they'll have a gap there?

I think I would actually expect the way robots should be checked is that the Checker interface gains a new `check_robots` method, which also takes a check, but explicitly only does the robots check.

Then when we're scheduling checks we'll decide if it's time to also execute the robots check and receive those results and produce them

src/config_store.rs
evanpurkhiser, line 1: unrelated formatting changes?

klochek: Unrelated, but breaking CI linting, somehow.

klochek:
Schema was correctly bumped in consumer, but not in uptime checker itself.  Reverting, then fixing (+ adding a test)

PR #411, opened 2025-10-29
fix(uptime): clean up some recv() usage and vec allocs

klochek:
In regions with very few uptime subscriptions (i.e. bringing up a new region,) we will block sending results to vector until we've accumulated enough results.  This bounds that buffering with an upper-bound of default 10 seconds.

### DIFF
--- a/src/app/config.rs
+++ b/src/app/config.rs
@@ -172,6 +172,9 @@ pub struct Config {
     /// The batch size to use for vector producer
     pub vector_batch_size: usize,
 
+    /// The timeout to control batch flushing for vector producer
+    pub vector_batch_timeout_secs: u64,
+
     /// The vector endpoint to send results to
     pub vector_endpoint: String,
 
@@ -270,6 +273,7 @@ impl Default for Config {
             checker_mode: CheckerMode::Reqwest,
             vector_batch_size: 10,
             vector_endpoint: "http://localhost:8020".to_owned(),
+            vector_batch_timeout_secs: 10,
             retry_vector_errors_forever: false,
             producer_mode: ProducerMode::Kafka,
             config_provider_redis_update_ms: 1000,
@@ -455,6 +459,7 @@ mod tests {
                         total_checkers: 1,
                         producer_mode: ProducerMode::Kafka,
                         vector_batch_size: 10,
+                        vector_batch_timeout_secs: 10,
                         vector_endpoint: "http://localhost:8020".to_owned(),
                         retry_vector_errors_forever: false,
                         failure_retries: 0,
@@ -528,6 +533,7 @@ mod tests {
                 ("UPTIME_CHECKER_WEBSERVER_PORT", "81"),
                 ("UPTIME_CHECKER_ASSERTION_COMPLEXITY", "120"),
                 ("UPTIME_CHECKER_MAX_ASSERTION_OPS", "15"),
+                ("UPTIME_CHECKER_VECTOR_BATCH_TIMEOUT_SECS", "11"),
             ],
             |config| {
                 assert_eq!(
@@ -574,6 +580,7 @@ mod tests {
                         total_checkers: 5,
                         producer_mode: ProducerMode::Kafka,
                         vector_batch_size: 10,
+                        vector_batch_timeout_secs: 11,
                         vector_endpoint: "http://localhost:8020".to_owned(),
                         retry_vector_errors_forever: false,
                         failure_retries: 2,
--- a/src/producer/vector_producer.rs
+++ b/src/producer/vector_producer.rs
@@ -72,12 +72,19 @@ impl VectorResultsProducer {
             // Avoid cloning this string over and over again.
             let region: &'static str = config.region.clone().leak();
             let mut batch = Vec::with_capacity(config.vector_batch_size);
+            let mut batch_element_count = 0;
 
             while receiver
-                .recv_many(&mut batch, config.vector_batch_size)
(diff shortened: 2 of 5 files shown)


wedamija (approved):

src/producer/vector_producer.rs
```diff
@@ -72,39 +77,61 @@ impl VectorResultsProducer {
...
-            while let Some(json) = receiver.recv().await {
-                let new_count = pending_items.fetch_sub(1, Ordering::Relaxed);
-                metrics::gauge!("producer.pending_items").set(new_count as f64);
-
-                batch.push(json);
-
-                if batch.len() < config.vector_batch_size {
+                if batch_element_remaining > 0 {
                     continue;
                 }
```
wedamija: I'm wondering if this `if` statement is useful anymore. Since we now use `recv_many`, we should either fetch `batch_element_remaining` elements, or hit a timeout. Either way, we just want to process the batch if it's not empty.

Just bring it up because at first this threw me off and I thought it was a bug, until I saw the `Err(_) => batch_element_remaining = 0` line above.

wedamija (approved):

src/producer/vector_producer.rs
klochek: `recv_many` can receive _up to_ the specified amount, but if there is less than that in the pipe, it'll just return everything.  It's only if the pipe is empty that it'll block.

src/app/config.rs
```diff
@@ -172,6 +172,9 @@ pub struct Config {
     /// The batch size to use for vector producer
     pub vector_batch_size: usize,
 
+    /// The batch size to use for vector producer
```
loewenheim: The doc string seems to belong to the field above.

loewenheim (approved):

src/producer/vector_producer.rs
```diff
@@ -72,39 +77,61 @@ impl VectorResultsProducer {
             // Avoid cloning this string over and over again.
             let region: &'static str = config.region.clone().leak();
             let mut batch = Vec::with_capacity(config.vector_batch_size);
+            let mut batch_element_remaining = config.vector_batch_size;
+            let timeout_duration: Duration = Duration::from_secs(config.vector_batch_timeout_secs);
+            let mut deadline = Instant::now()
+                .checked_add(timeout_duration)
+                .expect("timeout_duration is small");
```
Dav1dde: That happens when it is too large, right?

src/producer/vector_producer.rs
```diff
@@ -151,11 +178,14 @@ async fn send_batch(
...
-        .map(|s| s + "\n")
-        .collect();
+    let vec_size = batch.iter().fold(0, |acc, b| acc + b.len() + 1);
+    let batch_len = batch.len();
+
+    let mut body = Vec::with_capacity(vec_size);
+    for s in batch.drain(..) {
+        body.extend(s);
+        body.push(b'\n');
+    }
```
Dav1dde, line 153: There is a potential optimization here, you can express this as a stream and pass it to reqwest, meaning you don't actually need to eagerly build and concatenate the body.

Dav1dde (approved):

src/producer/vector_producer.rs
```diff
@@ -72,39 +77,61 @@ impl VectorResultsProducer {
...
             let mut batch = Vec::with_capacity(config.vector_batch_size);
+            let mut batch_element_remaining = config.vector_batch_size;
+            let timeout_duration: Duration = Duration::from_secs(config.vector_batch_timeout_secs);
+            let mut deadline = Instant::now()
+                .checked_add(timeout_duration)
+                .expect("timeout_duration is small");
+
+            loop {
+                let receive_fut = receiver.recv_many(&mut batch, batch_element_remaining);
+                let result = timeout_at(deadline.into(), receive_fut).await;
```
jjbayer: nit: You can convert `deadline` to a tokio `Instant` outside the loop.

src/producer/vector_producer.rs
```diff
@@ -72,39 +77,61 @@ impl VectorResultsProducer {
             // Avoid cloning this string over and over again.
             let region: &'static str = config.region.clone().leak();
             let mut batch = Vec::with_capacity(config.vector_batch_size);
+            let mut batch_element_remaining = config.vector_batch_size;
+            let timeout_duration: Duration = Duration::from_secs(config.vector_batch_timeout_secs);
+            let mut deadline = Instant::now()
+                .checked_add(timeout_duration)
+                .expect("timeout_duration is small");
+
+            loop {
```
jjbayer: You might be able to build this mechanism with tokio types: `UnboundedReceiver` can be wrapped in an [`UnboundedReceiverStream`](https://docs.rs/tokio-stream/0.1.18/tokio_stream/wrappers/struct.UnboundedReceiverStream.html), which supports getting batches of a certain size with a timeout: https://docs.rs/tokio-stream/0.1.18/tokio_stream/trait.StreamExt.html#method.chunks_timeout

klochek: lovely!  much, much nicer

src/producer/vector_producer.rs
klochek: This _almost_ works, except that because we need to retry, we have to store the body bytes (and clone them,) and so we're right back to where we started :(

src/producer/vector_producer.rs
klochek: The convention I've been using in uptime-checker is that, for `expect()`s, the string is documentation for why the expect should always be okay, rather than documentation indicating what went wrong.

PR #446, opened 2025-12-08
Cargo.toml - enable TLS on redis crate

zmarta-erik:
Hello,

We use a self hosted version of Sentry. In our setup we use Redis with TLS encryption (AWS Elasticache) which works well with other components of Sentry, but not uptime-checker.

Enabling support for TLS on the redis crate fixes this issue. For other methods of self hosting Redis, one might wish to enable 'tls-rustls-insecure' as well (to allow self signed certificates).

I did take a look at previous PRs and issues and I did not spot any previous mention of this, my bad if I overlooked something!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,12 @@ metrics-exporter-statsd = "0.8.0"
 metrics = "0.23.0"
 futures = "0.3.30"
 tokio-stream = "0.1.15"
-redis = { version = "0.30.0", features = ["cluster-async", "tokio-comp"] }
+redis = { version = "0.30.0", features = [
+  "cluster-async",
+  "tokio-comp",
+  "tls-rustls",
+  "tokio-rustls-comp"
+] }
 ipnet = "2.11.0"
 hickory-resolver = "0.24.1"
 tokio-rustls = "0.26.1"


evanpurkhiser (approved):

SmoshySmosh:
@evanpurkhiser @aldy505 Any chance we could land this? I'm migrating my self-hosted stack to AWS ElastiCache Valkey Serverless, which requires TLS — without these features the redis crate panics at startup on `rediss://` URLs (`can't connect with TLS, the feature is not enabled`).

I'm building against this branch locally to validate end-to-end and will follow up here with results.

SmoshySmosh:
The changes in this PR alone are not enough, but I have a confirmed change working end-to-end against AWS ElastiCache Valkey Serverless (TLS-required, cluster-mode endpoint). Two additions on top of the current branch were needed to get rediss:// URLs actually connecting:

1. `rustls` feature. Rustls 0.23 no longer auto-selects a crypto backend, so building with `tls-rustls` compiles but there's no provider to hand off to. Adding `features = ["ring"]` gives rustls a backend to expose at `rustls::crypto::ring`.

2. Process-wide provider install. Even with the feature enabled, rustls panics on first handshake (`no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point`) unless the application installs one at startup. This is the only place it can be installed. The choice is process-global, so libraries like redis deliberately don't do it themselves.

I also trimmed the `redis` feature list. `tokio-rustls-comp` transitively enables `tokio-comp` + `tls-rustls`, and `tls-rustls-webpki-roots` also enables `tls-rustls`, so both can be dropped. Verified this produces a byte-for-byte identical binary to the fuller list.

Changes on top of this PR:

**Cargo.toml**
```
redis = { version = "0.30.0", features = ["cluster-async", "tokio-rustls-comp", "tls-rustls-webpki-roots"] }
rustls = { version = "0.23.21", features = ["ring"] }
```

**src/main.rs**
```
 pub fn main() {
    // Rustls 0.23 no longer auto-selects a crypto provider; libraries that
    // use rustls (e.g. `redis` with tls-rustls) panic on first TLS handshake
    // without this. Installing here — the only spot that can, since the
    // choice is process-wide.
    rustls::crypto::ring::default_provider()
        .install_default()
        .expect("failed to install rustls crypto provider");
```         

Adjust the `redis` feature list to match whatever set you'd prefer. The minimum for `rediss://` cluster + tokio is `cluster-async` + `tokio-rustls-comp`. I kept `tls-rustls-webpki-roots` because ElastiCache endpoints validated more reliably against the bundled Mozilla roots than against Alpine's `ca-certificates` package on the runtime image; feel free to drop it if you'd rather rely on `rustls-native-certs`.

aldy505:
@SmoshySmosh working it here https://github.com/getsentry/uptime-checker/pull/506

SmoshySmosh:
@aldy505 Thank you! I was thinking about submitting a PR but I can't use github at work 😅.

PR #498, opened 2026-06-16
fix(gocd): exclude s4s2 from uptime-checker pipedream

mchen-sentry:
GoCD has been rejecting the uptime-checker config repo since the 2026-05-27 region rename. The generated `deploy-uptime-checker-us2` and `-s4s2` pipelines come out with empty `deploy-primary`, `deploy-canary`, and `scale-down-canary` stages, and GoCD won't accept a stage with no jobs, so the parse fails and the whole repo's config stays frozen on last-known-good.

Those stages generate one job per entry in `region_pops[region]`, and both `us2` and `s4s2` map to `[]` since uptime-checker has no POPs in those regions yet. The ops-side pipedream already excludes both (done in the same rename, getsentry/ops#20845), but this repo's copy was missed even though the libsonnet says to keep the two in sync.

Adding `us2` and `s4s2` to `exclude_regions` stops pipedream from generating those empty pipelines. When real POPs land in those regions we'll populate `region_pops` and drop them from the exclude list at the same time. The gocd-jsonnet validation in CI confirms the config parses.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4736,7 +4736,7 @@ checksum = "8ecb6da28b8a351d773b68d5825ac39017e680750f980f3a1a85cd8dd28a47c1"
 
 [[package]]
 name = "uptime-checker"
-version = "26.5.2"
+version = "26.6.0"
 dependencies = [
  "anyhow",
  "associative-cache",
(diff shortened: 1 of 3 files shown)


BojanOro:
Added US2 since we deployed uptime there :)

klochek (approved):

PR #499, opened 2026-06-24
fix(uptime): Handle redis broken pipes better

klochek:
We just hit this in prod (and self-hosted folks noticed it a few weeks ago, coincidentally.)  If our connection to redis is broken, we don't try to make a new connection, leading to log spam and no longer retrieving updated configs.  This has the monitor_configs function handle redis errors (we still log, but now we sleep and retry the entire monitoring loop again.)

### DIFF
--- a/src/check_config_provider/redis_config_provider.rs
+++ b/src/check_config_provider/redis_config_provider.rs
@@ -131,7 +131,7 @@ impl RedisConfigProvider {
                 if let Err(err) = result {
                     tracing::error!(?err, "redis_config_provider.monitor_configs");
 
-                    // For reddis errors, wait and retry.
+                    // For redis errors, wait and retry.
                     tokio::time::sleep(Duration::from_secs(5)).await;
                 }
             }
--- a/src/check_config_provider/redis_config_provider.rs
+++ b/src/check_config_provider/redis_config_provider.rs
@@ -220,11 +220,7 @@ impl RedisConfigProvider {
         shutdown: CancellationToken,
         region: &'static str,
     ) -> Result<(), RedisError> {
-        let mut ops = self
-            .redis
-            .get_async_connection()
-            .await
-            .expect("Redis should be available");
+        let mut ops = self.redis.get_async_connection().await?;
         let mut interval = interval(self.check_interval);
 
         while !shutdown.is_cancelled() {
--- a/src/check_config_provider/redis_config_provider.rs
+++ b/src/check_config_provider/redis_config_provider.rs
@@ -2,6 +2,7 @@ use crate::{
     app::config::Config, manager::Manager, redis::RedisClient, types::check_config::CheckConfig,
 };
 use anyhow::Result;
+use redis::RedisError;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/check_config_provider/redis_config_provider.rs
```diff
@@ -122,8 +123,18 @@ impl RedisConfigProvider {
...
+            while !shutdown.is_cancelled() {
+                let result = self
+                    .monitor_updates(manager.clone(), &partitions, shutdown.clone(), region)
+                    .await;
+
+                if let Err(err) = result {
+                    tracing::error!(?err, "redis_config_provider.monitor_configs");
+
+                    // For reddis errors, wait and retry.
+                    tokio::time::sleep(Duration::from_secs(5)).await;
```
Dav1dde: Does uptime have an easy config system which would allow making such values configurable or a list of documented constants?

src/check_config_provider/redis_config_provider.rs
```diff
@@ -205,12 +216,8 @@ impl RedisConfigProvider {
...
         shutdown: CancellationToken,
         region: &'static str,
-    ) {
-        let mut ops = self
-            .redis
-            .get_async_connection()
-            .await
-            .expect("Redis should be available");
+    ) -> Result<(), RedisError> {
+        let mut ops = self.redis.get_async_connection().await?;
```
Dav1dde, line 220: Just me asking questions: Is this a somewhat hot path? Seems like this will create a new connection everytime instead of pooling, which may be problematic.

Dav1dde (approved):

src/check_config_provider/redis_config_provider.rs
klochek: We do actually have that (there's a giant config here https://github.com/getsentry/uptime-checker/blob/main/src/app/config.rs that can be loaded in via envars); I opted to not do that here because this feels more like a hack than something that deserves a config field.  Relatedly: we're actually quite out-of-date with the rust redis library, and if we upgraded, we'd get access to a managed connection pool mechanism that implements exponential backoff.  This PR fixes the proximal problem, but a nice to have would be to upgrade redis.

src/check_config_provider/redis_config_provider.rs
klochek: So, the loop below these lines is where we spend all of our time (asking redis for updates,) if we ever leave it, it's because we're either (a) shutting down, or (b) connection error, in which case we activate the retry logic here.  Once in the loop, we re-used the same connection to redis.

src/check_config_provider/redis_config_provider.rs
```diff
@@ -122,8 +123,18 @@ impl RedisConfigProvider {
...
-                .await;
+            while !shutdown.is_cancelled() {
+                let result = self
+                    .monitor_updates(manager.clone(), &partitions, shutdown.clone(), region)
+                    .await;
+
+                if let Err(err) = result {
+                    tracing::error!(?err, "redis_config_provider.monitor_configs");
+
+                    // For reddis errors, wait and retry.
```
jjbayer:
```suggestion
                    // For redis errors, wait and retry.
```

jjbayer (approved):

PR #502, opened 2026-06-30
chore(gocd): bump gocd-jsonnet to v3.0.6

mchen-sentry:
Bumps gocd-jsonnet to v3.0.6 for the single-region gate-skip fix in getsentry/gocd-jsonnet#97.

PR #503, opened 2026-06-30
fix(gocd): re-enable us2 uptime-checker deploys

mchen-sentry:
#502 excluded `us2` from the uptime-checker pipedream, but that exclude commit was written back when `us2`'s pops were a copy-paste of the `us` clusters, which is no longer the case: getsentry/uptime-checker#501 

Undoing the change here.


Dav1dde:
@mchen-sentry thanks! Gonna wait on @BojanOro / @klochek to confirm that everything is in order for that pipeline.

Dav1dde (approved):
Actually nvm, you already linked the fix PR.

PR #512, opened 2026-08-27
docs(redis): correct the update key's structure and commands

ldelvoye:
Fixes INFRENG-486

### DIFF
--- a/src/redis.rs
+++ b/src/redis.rs
@@ -68,8 +68,7 @@ impl RedisOperations {
             .expect("must be in read-write mode to access");
 
         let mut pipe = redis::pipe();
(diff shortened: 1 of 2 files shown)


src/redis.rs
```diff
@@ -68,15 +68,15 @@ impl RedisOperations {
             .expect("must be in read-write mode to access");
 
         let mut pipe = redis::pipe();
-        // We fetch all updates from the list and then delete the key. We do this
-        // atomically so that there isn't any chance of a race
+        // HVALS and DEL the update hash in one MULTI, so an update written between the
```
klochek: I think the old comment is a bit clearer (if less precise.)

klochek (approved):

